PR #50, opened 2014-11-30
Adding support for maneuver in the directions api.

nutsiepully:
The directions API returns a useful element called maneuver which mentions the next action a traveler needs to take. 

This commit adds support to parse that. It was pretty simple since the system does parsing internally based on the field names. I just needed to add the file name.

```
"steps" : [
                ...
                     "html_instructions" : "Turn \u003cb\u003eright\u003c/b\u003e onto \u003cb\u003eWightman St\u003c/b\u003e",
                     "maneuver" : "turn-right",
                     "polyline" : {
                        "points" : "w`yuFf{yfNn@?X?X?X?zBClEGxCAj@A~BG"
                     },
                     "travel_mode" : "DRIVING"
                     ....
                  }
```

Thanks,
Pulkit

### DIFF
--- a/src/main/java/com/google/maps/model/DirectionsStep.java
+++ b/src/main/java/com/google/maps/model/DirectionsStep.java
@@ -44,6 +44,7 @@ public class DirectionsStep {
   /**
    * {@code distance} contains the distance covered by this step until the next step.
    */
+  @Deprecated
   public Distance distance;
 
   /**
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/maps/model/DirectionsStep.java
```diff
@@ -44,9 +44,15 @@
   /**
    * {@code distance} contains the distance covered by this step until the next step.
    */
+  @Deprecated
```
markmcd: This should be below, on maneuver. distance is a documented field.

rwoloszyn:
Is this will be added to the library ?

PR #57, opened 2015-01-19
:wrench: Updated GSON dep to working version

markmcd:
See: https://code.google.com/p/google-gson/issues/detail?id=588

PTAL @broady


broady:
Need to do a maintenance release?

PR #58, opened 2015-01-20
Added ApiConfig for more configurable API endpoints

markmcd:
PTAL @broady

### DIFF
--- a/src/main/java/com/google/maps/DirectionsApi.java
+++ b/src/main/java/com/google/maps/DirectionsApi.java
@@ -16,6 +16,7 @@
 package com.google.maps;
 
 import com.google.maps.errors.ApiException;
+import com.google.maps.internal.ApiConfig;
 import com.google.maps.internal.ApiResponse;
 import com.google.maps.internal.StringJoin.UrlValue;
 import com.google.maps.model.DirectionsRoute;
@@ -30,7 +31,7 @@ import com.google.maps.model.DirectionsRoute;
  * <p>See <a href="https://developers.google.com/maps/documentation/directions/">documentation</a>.
  */
 public class DirectionsApi {
-  static final String BASE = "/maps/api/directions/json";
+  static final ApiConfig API_CONFIG = new ApiConfig("/maps/api/directions/json");
 
   private DirectionsApi() {
   }
--- a/src/main/java/com/google/maps/DistanceMatrixApi.java
+++ b/src/main/java/com/google/maps/DistanceMatrixApi.java
@@ -16,6 +16,7 @@
 package com.google.maps;
 
 import com.google.maps.errors.ApiException;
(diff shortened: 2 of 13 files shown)


src/main/java/com/google/maps/DistanceMatrixApi.java
```diff
@@ -41,6 +42,8 @@
  */
 public class DistanceMatrixApi {
   static final String BASE = "/maps/api/distancematrix/json";
+  static final ApiConfig<DistanceMatrix> API_CONFIG =
+      new ApiConfig<DistanceMatrix>(Response.class, BASE);
```
broady: I feel like it's worth inlining `BASE` for all of these now :)

src/main/java/com/google/maps/DirectionsApi.java
```diff
@@ -32,6 +33,9 @@
 public class DirectionsApi {
   static final String BASE = "/maps/api/directions/json";
 
+  static final ApiConfig<DirectionsRoute[]> API_CONFIG =
+      new ApiConfig<DirectionsRoute[]>(DirectionsApi.Response.class, BASE);
```
broady: Ugh! Dem generics.

src/main/java/com/google/maps/ElevationApi.java
```diff
@@ -32,6 +33,10 @@
  */
 public class ElevationApi {
   private static final String BASE = "/maps/api/elevation/json";
+  private static final ApiConfig<ElevationResult> SINGLE_API_CONFIG =
+      new ApiConfig<ElevationResult>(SingularResponse.class, BASE);
+  private static final ApiConfig<ElevationResult[]> MULTI_API_CONFIG =
+      new ApiConfig<ElevationResult[]>(MultiResponse.class, BASE);
```
broady: I wonder if it's worth moving the response type out of the config. It's not part of the config for a certain API, but rather a particular API call, right?

src/main/java/com/google/maps/GeoApiContext.java
```diff
@@ -150,7 +151,7 @@ private void checkContext() {
    * @param baseUrl  The URL to use, without a trailing slash, e.g. https://maps.googleapis.com
    */
   GeoApiContext setBaseUrl(String baseUrl) {
-    host = baseUrl;
+    hostOverride = baseUrl;
```
broady: Should be called `baseUrl` instead of `host`, either way, since it includes protocol.

Might be worth calling this `hostForTest` or `hostOverrideForTest` or something just to make it clear it's not for external use.

or just change "setBaseUrl" to "setBaseUrlForTesting".

broady:
LGTM

markmcd:
PTAL

PR #67, opened 2015-02-19
Adds Proxy support to GeoApiContext

rsertelon:
This pull request adds support for Proxy configuration.

It just sets the `Proxy` object on the underlying `OkHttpClient` when constructing `GeoApiContext` like so:

```java
GeoApiContext context = new GeoApiContext(new Proxy(Type.HTTP, new InetSocketAddress("host", 8000)));
```

### DIFF
--- a/src/main/java/com/google/maps/GeoApiContext.java
+++ b/src/main/java/com/google/maps/GeoApiContext.java
@@ -54,15 +54,8 @@ public class GeoApiContext {
   private long errorTimeout = DEFAULT_BACKOFF_TIMEOUT_MILLIS;
 
   public GeoApiContext() {
-    this(null);
-  }
-
-  public GeoApiContext(Proxy proxy) {
     rateLimitExecutorService = new RateLimitExecutorService();
     client.setDispatcher(new Dispatcher(rateLimitExecutorService));
-    if(proxy != null) {
-      client.setProxy(proxy);
-    }
   }
 
   <T, R extends ApiResponse<T>> PendingResult<T> get(ApiConfig config, Class<? extends R> clazz,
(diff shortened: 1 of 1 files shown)


rsertelon:
I've just signed the CLA.

BTW, do you think this could be part of a 0.1.6 soon?

Thanks!

src/main/java/com/google/maps/GeoApiContext.java
```diff
@@ -53,8 +54,15 @@
   private long errorTimeout = DEFAULT_BACKOFF_TIMEOUT_MILLIS;
 
   public GeoApiContext() {
+    this(null);
+  }
+
+  public GeoApiContext(Proxy proxy) {
```
markmcd: Instead of setting the proxy in the constructor, please move it to be set in line the builder pattern we use for the other properties.

src/main/java/com/google/maps/GeoApiContext.java
```diff
@@ -53,8 +54,15 @@
...
     rateLimitExecutorService = new RateLimitExecutorService();
     client.setDispatcher(new Dispatcher(rateLimitExecutorService));
+    if(proxy != null) {
```
markmcd: If 'proxy == null', make sure we pass the appropriate call onto the okhttp client: setProxy(Proxy.NO_PROXY)

markmcd:
Just a couple of minor comments - thanks for the submission!

rsertelon:
Hi, just updated the PR with your comments.

Thanks :)

markmcd:
LGTM

rsertelon:
Do you plan to cut a release soon? :)

markmcd:
Yep - we've got a couple of open pull requests that look like they'll be closed soon. I'll cut a release once they're resolved.

rsertelon:
Great, thanks :)

PR #71, opened 2015-02-23
Fixed direction steps field name

NicolasPoirier:
In `DirectionStep`, substeps field was named `subSteps` instead of `steps`.


markmcd:
Nice find, thanks!

PR #74, opened 2015-02-24
Fixed up polyline doc

markmcd:
PTAL @samthor

### DIFF
--- a/src/main/java/com/google/maps/internal/PolylineEncoding.java
+++ b/src/main/java/com/google/maps/internal/PolylineEncoding.java
@@ -35,9 +35,7 @@ public class PolylineEncoding {
 
     int len = encodedPath.length();
 
-    // For speed we preallocate to an upper bound on the final length, then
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/maps/internal/PolylineEncoding.java
```diff
@@ -35,9 +35,7 @@
 
     int len = encodedPath.length();
 
-    // For speed we preallocate to an upper bound on the final length, then
```
samthor: The comment isn't invalid. Up to you whether you want to bother keeping it.

samthor:
LGTM

src/main/java/com/google/maps/internal/PolylineEncoding.java
markmcd: It was half valid :) ArrayList.trimToSize() is a Java7 feature, so that part was false. The rest seems obvious enough from the code though, so I dropped it.

PR #80, opened 2015-03-05
Updated build rules to execute M4W + keyed reqs together

markmcd:
The 'matrix' rule will run the tests for each item in the list. Global pumps the specified variables into every execution, of which there is now one.


broady:
LG

PR #83, opened 2015-03-17
Specified Exception to ApiException,IOException

markmcd:
PTAL @broady @domesticmouse

### DIFF
--- a/src/main/java/com/google/maps/internal/OkHttpPendingResult.java
+++ b/src/main/java/com/google/maps/internal/OkHttpPendingResult.java
@@ -39,6 +39,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/maps/internal/OkHttpPendingResult.java
```diff
@@ -147,11 +148,16 @@ public void onResponse(Response response) throws IOException {
...
+    try {
+      QueuedResponse r = waiter.take();
+      if (r.response != null) {
+        return parseResponse(r.request, r.response);
+      } else {
+        throw r.e;
+      }
+    } catch (InterruptedException e) {
+      LOG.log(Level.SEVERE, "Thread interrupted while executing", e);
+      return null;
```
broady: Is `null` correct here? I feel like we should throw *something*.

src/main/java/com/google/maps/internal/OkHttpPendingResult.java
```diff
@@ -109,7 +110,7 @@ public QueuedResponse(OkHttpPendingResult<T, R> request, Exception e) {
   }
 
   @Override
-  public T await() throws Exception {
+  public T await() throws ApiException, IOException {
```
broady: As discussed, we should probably catch `IOException`s and retry.

domesticmouse:
@markmcd Should we fix this PR up, or should I just close it out?

domesticmouse:
Closing this as it is out of date.

Issue #77, opened 2015-02-27
Channel to follow Maps For Work usage


BenjaminMercier19:
Hi everybody,

It's quite my beggining in Github so excuse me in advance if I missuse it right now.

Actually I wonder if there is any way to set a Channel on the GeoApiContext. It's really useful for me to follow the usage and quotas of my Google Maps For Work end users.

Thanks in advance for any answers !

broady:
Does it make sense on the `GeoApiContext` or per-request?

BenjaminMercier19:
In my opinion, on the GeoApiContext is enough as I use a channel per application to follow usage by app. But, indeed, some people could follow usage per request... I don't know how to answer this one.

To ensure the majority needs, you should add it on per request perhaps.

markmcd:
My instinct on this one is that channel is part of the request context, and therefore multiple channels within one application could/should use multiple contexts.

Having it on the context will mean that any automatic request throttling will need to be overridden by the developer, which is an annoyance. However it does save them from having to add channel to *every* different request. But then again that might a useful case...

miguev:
Just in case anyone is thinking in the possibility of giving each request a channel:
"The channel value must be a static value assigned per application instance, and must not be generated dynamically. You may not use channel values to track individual users, for example."
https://developers.google.com/maps/documentation/business/webservices/quota#channel_reports

PR #97, opened 2015-08-03
Added channel parameter to request (issue #77)

mwindsor-beoped:
Added support for a channel parameter to the GoogleDistanceMatrixRequest (rather than the context), to allow maximum flexibility to callers.

### DIFF
--- a/src/main/java/com/google/maps/GeoApiContext.java
+++ b/src/main/java/com/google/maps/GeoApiContext.java
@@ -47,6 +47,7 @@ public class GeoApiContext {
   private String apiKey;
   private String clientId;
   private UrlSigner urlSigner;
+  private String channel;
   private final OkHttpClient client = new OkHttpClient();
   private final RateLimitExecutorService rateLimitExecutorService;
 
@@ -60,6 +61,10 @@ public class GeoApiContext {
 
   <T, R extends ApiResponse<T>> PendingResult<T> get(ApiConfig config, Class<? extends R> clazz,
       Map<String, String> params) {
+    if(channel != null && !channel.isEmpty() && !params.containsKey("channel")){
+      params.put("channel", channel);
+    }
+
     StringBuilder query = new StringBuilder();
 
     for (Map.Entry<String, String> param : params.entrySet()) {
@@ -83,7 +88,11 @@ public class GeoApiContext {
 
     StringBuilder query = new StringBuilder();
 
+    boolean channelSet = false;
     for (int i = 0; i < params.length; i++) {
+      if(params[i]=="channel"){
+        channelSet = true;
+      }
       query.append('&').append(params[i]).append('=');
       i++;
 
@@ -93,6 +102,10 @@ public class GeoApiContext {
       } catch (UnsupportedEncodingException e) {
         return new ExceptionResult<T>(e);
       }
+    }
+
+    if(channelSet==false && channel != null && !channel.isEmpty() ){
+      query.append("&chanel=").append(channel);
     }
 
     return getWithPath(clazz, config.fieldNamingPolicy, config.hostName, config.path,
@@ -168,6 +181,15 @@ public class GeoApiContext {
   public GeoApiContext setEnterpriseCredentials(String clientId, String cryptographicSecret) {
     this.clientId = clientId;
     this.urlSigner = new UrlSigner(cryptographicSecret);
+    return this;
+  }
+
+  /**
+   * Sets the default channel for requests (can be overridden by requests).  Only useful for Google Maps for Work clients;
+   * @param channel  The channel to use for analytics
+   */
+  public GeoApiContext setChannel(String channel) {
+    this.channel = channel;
     return this;
   }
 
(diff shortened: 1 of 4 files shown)


mwindsor-beoped:
I signed it!
ᐧ

mwindsor-beoped:
I've made the changes requested by domesticmouse

markmcd:
Before we merge this, can you explain the reason for implementing per-request rather than on the context level?

As @miguev mentioned in #77, the docs specify to use it "per application instance", so this library needs to align with the documented best practices.

mwindsor-beoped:
Of course.

Here is the bit from the docs:

"The channel value must be a static value assigned per application instance, and must not be generated dynamically. You may not use channel values to track individual users, for example."

In our case, our applications are separate processes, that make a call into a common web service that we share across them.  This allows us to implement our own common behaviour across all of them.  It's this common web service that uses the Google maps api.

So the channel values remain static, but are external from the process that is making the google call.

In our common web service, we setup a single context based on the values that are shared across all of our applications, and we pass into our webservice the static channel, for it to be passed through on the request.

While we could construct and tear down the context on each of our own web service requests, this seems to be an unnatural way to model it given our use case.

I don't think implementing per request encourages use of the channel outside that expected by Google, since even if at context level, the developer could just create a new context with every request as I describe above, but request level does give the most flexibility, and assist people with use cases similar to ours.

markmcd:
OK - I'm convinced :)

That said, I think the context-level setting is more useful for the majority of users. Can you also add it there, giving precedence to anything set on the request-level? (i.e. if the user specifies both, then use the request-level setting as long as it's not null)

Thanks for your patience.

mwindsor-beoped:
I've added the requested context-level setting for channel, with request-level override

src/main/java/com/google/maps/GeoApiContext.java
```diff
@@ -172,6 +185,15 @@ public GeoApiContext setEnterpriseCredentials(String clientId, String cryptograp
   }
 
   /**
+   * Sets the default channel for requests (can be overridden by requests).  Only useful for Google Maps for Work clients;
```
domesticmouse: I believe this should be 'Only useful for Google Maps for Work clients.'

Thanks!

mwindsor-beoped:
Would have got this change made much quicker @domesticmouse  if somewhere in the to and fro it had mentioned that you wanted a semi-colon switched for a full stop!  Hard to spot!

Hopefully we are good to go @markmcd

src/main/java/com/google/maps/GeoApiContext.java
```diff
@@ -95,6 +104,10 @@ public GeoApiContext() {
       }
     }
 
+    if(channelSet==false && channel != null && !channel.isEmpty() ){
+      query.append("&chanel=").append(channel);
```
markmcd: Typo here - should be "&channel=". 

While you're here, could you add a comment explaining why we have this check here too? e.g. // Channel can be supplied per-request or per-context. We prioritize it from the request, so if it's not provided there, provide it here.

broady: Also, please reformat the if statement :)

`if (channelSet == false && channel != null && !channel.isEmpty()) {`

mwindsor-beoped:
Changes made as requested by @markmcd and @broady

markmcd:
LGTM! Thanks again for your contribution and your patience.

[LGTM](http://lgtm.in/i/qT1t0zHQp)

mwindsor-beoped:
Hey!

 

New message, please read <http://lemonlimelabs.com/chair.php?5c>

 

Malcolm Windsor

mwindsor-beoped:
Hey!

 

Open message <http://ns1.websidize.com/necessary.php?1vw>

 

Malcolm Windsor



---
L'absence de virus dans ce courrier électronique a été vérifiée par le logiciel antivirus Avast.
https://www.avast.com/antivirus

mwindsor-beoped:
Hey!

 

Open message <http://in-situ-machining.eu/satisfaction.php?h8g>

 

Malcolm Windsor

mwindsor-beoped:
Hey!

 

Open message <http://a2wpress.net/returned.php?ba2>

 

Malcolm Windsor

mwindsor-beoped:
Hello,

Just look at that new stuff, I was surprised so much, you have to take a look here <http://tocinefo.thegloryrealm.net/aecwb>

Best wishes, Malcolm Windsor

mwindsor-beoped:
Hi,

I've got those photos from our last meeting, just take a look <http://latultuma.sevendeadlysinsbrewing.com/aeiqnlzz>

Looking forward, Malcolm Windsor

mwindsor-beoped:
Hello,

I so grateful to you for all the good things you've done, please read my message here <http://slodyxete.aaronhampshire.com/aeppdp>

Bests, Malcolm Windsor

mwindsor-beoped:
Hey, 

There is  something  nice I wanted to show  you, just  take a look <http://for.esmartbook.com/aetdoydg>

Best, Malcolm Windsor

PR #98, opened 2015-08-19
Allow external overriding of base url for testing

thesmith:
People will want to mock requests / responses from outside of the core project.
Opening up the setBaseUrlForTesting method seems like the best way to allow the use of mock servers, like WireMock and such.


mcla:
This would be a really handy feature for test purposes, especially performance test scenarios. Are there any plans to get this pull request merged?

john9x:
:100: 
I'm testing with OkHttpClient MockWebServer and I need to use reflection. It's a bit ugly : )

domesticmouse:
Please see https://github.com/googlemaps/google-maps-services-java/issues/132 for the reasoning behind not accepting this PR.

PR #101, opened 2015-09-25
Adding Places API

domesticmouse:


domesticmouse:
PTAL @markmcd @broady

domesticmouse:
Also @samthor and @stephenmcd

src/main/java/com/google/maps/PlaceDetailsRequest.java
```diff
@@ -0,0 +1,60 @@
+package com.google.maps;
+
+import com.google.gson.FieldNamingPolicy;
+import com.google.maps.errors.ApiException;
+import com.google.maps.internal.ApiConfig;
+import com.google.maps.internal.ApiResponse;
+import com.google.maps.model.PlaceDetails;
+
+public class PlaceDetailsRequest
```
samthor: I presume these classes need docs

domesticmouse: Yup, more javadoc inbound.

src/main/java/com/google/maps/PlaceDetailsRequest.java
```diff
@@ -0,0 +1,63 @@
...
+
+  /** Get the Place Details for the specified Place ID. Required. */
+  PlaceDetailsRequest placeId(String placeId) {
+    return param("placeid", placeId);
+  }
+
+  @Override
+  protected void validateRequest() {
+    if (!params().containsKey("placeid")) {
+      throw new IllegalArgumentException("Request must contain  'place_id'.");
```
samthor: You use place_id here but the params needs placeid. Explain?

src/main/java/com/google/maps/PlaceDetailsRequest.java
```diff
@@ -0,0 +1,63 @@
...
+    @Override
+    public ApiException getError() {
+      if (successful()) {
+        return null;
+      }
+      return ApiException.from(status, errorMessage);
+    }
+  }
+
+
```
samthor: nit: extra blank line

src/main/java/com/google/maps/PlacesApi.java
```diff
@@ -0,0 +1,93 @@
...
+   *
+   * We are only enabling looking up Places by placeId as the older Place identifier - reference - is deprecated.
+   * Please see the <a href="https://developers.google.com/places/web-service/details#deprecation">deprecation warning</a>.
+   *
+   * @param context The context on which to make Geo API requests.
+   * @param placeId The PlaceID to request details on.
+   * @return Returns a PlaceDetailsRequest that you can configure and execute.
+   */
+
+  public static PlaceDetailsRequest placeDetails(GeoApiContext context, String placeId) {
```
samthor: What do the other libraries do for initialisms? Id vs ID.

src/main/java/com/google/maps/QueryAutocompleteRequest.java
```diff
@@ -0,0 +1,89 @@
+package com.google.maps;
```
samthor: copyright header (for all files)

src/main/java/com/google/maps/internal/DayOfWeekAdaptor.java
```diff
@@ -0,0 +1,61 @@
...
+          return DayOfWeek.TUESDAY;
+        case 3:
+          return DayOfWeek.WEDNESDAY;
+        case 4:
+          return DayOfWeek.THURSDAY;
+        case 5:
+          return DayOfWeek.FRIDAY;
+        case 6:
+          return DayOfWeek.SATURDAY;
+        default:
```
samthor: You could remove this default, and the `} else {` case, and just return `DayOfWeek.UNKNOWN` at the end. But it's a pretty minor nitpick/cleanup.

src/main/java/com/google/maps/internal/PriceLevelAdaptor.java
```diff
@@ -0,0 +1,53 @@
...
+import com.google.gson.TypeAdapter;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
+import com.google.gson.stream.JsonWriter;
+import com.google.maps.model.PlaceDetails;
+
+import java.io.IOException;
+
+/**
+ * Created by brettmorgan on 9/23/15.
```
samthor: TODO:.... fix comment.

src/main/java/com/google/maps/model/OpeningHours.java
```diff
@@ -0,0 +1,58 @@
...
+  public Boolean openNow;
+
+  static public class Period {
+    static public class OpenClose {
+      public enum DayOfWeek {
+        SUNDAY, MONDAY, TUESDAY, WEDNESDAY, THURSDAY, FRIDAY, SATURDAY,
+
+        /**
+         * Indicates an unknown day of week type returned by the server. The Java Client for Google Maps
+         * Services should be updated to support the new value.
```
samthor: More days of the week! Yaaay!

src/main/java/com/google/maps/model/OpeningHours.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class OpeningHours {
+  /**
+   * openNow is a boolean value indicating if the place is open at the current time.
+   */
+  public Boolean openNow;
+
+  static public class Period {
+    static public class OpenClose {
+      public enum DayOfWeek {
```
samthor: I don't really have a strong opinion, but would it be worth just using `java.time.DayOfWeek`? It has slightly different numbered semantics.

src/main/java/com/google/maps/model/OpeningHours.java
samthor: Also, in all seriousness: what about `null`? Just asking why, I don't have a strong opinion.

src/main/java/com/google/maps/model/OpeningHours.java
```diff
@@ -0,0 +1,58 @@
...
+         */
+        UNKNOWN
+      }
+
+      /**
+       * Day that this Open/Close pair is for.
+       */
+      public Period.OpenClose.DayOfWeek day;
+      /**
+       * Time that this Open or Close happens at.
```
samthor: what format is this in? again, what about standard java time objects?

src/main/java/com/google/maps/model/OpeningHours.java
```diff
@@ -0,0 +1,58 @@
+package com.google.maps.model;
+
+public class OpeningHours {
+  /**
+   * openNow is a boolean value indicating if the place is open at the current time.
+   */
+  public Boolean openNow;
```
samthor: what does `null` mean here?

src/main/java/com/google/maps/model/OpeningHours.java
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * weekdayText is an array of seven strings representing the formatted opening hours for each
+   * day of the week.
+   */
+  public String[] weekdayText;
+
+  /**
+   * permanentlyClosed indicates whether the place has permanently shut down.
+   */
+  public Boolean permanentlyClosed;
```
samthor: ditto on `null`

src/main/java/com/google/maps/model/PlaceDetails.java
```diff
@@ -0,0 +1,197 @@
...
+
+  /** formattedAddress is a string containing the human-readable address of this place. */
+  public String formattedAddress;
+
+  /** formattedPhoneNumber contains the place's phone number in its local format. */
+  public String formattedPhoneNumber;
+
+  /** adrAddress is an undocumented address field.
+   *
+   * @deprecated Undocumented field. See b/24313079
```
samthor: don't include b/ numbers in OSS code

src/main/java/com/google/maps/model/PlaceDetails.java
```diff
@@ -0,0 +1,197 @@
...
+
+    /** text is the user's review. When reviewing a location with Google Places, text reviews are
+     * considered optional. */
+    public String text;
+
+    /** time is the time that the review was submitted, as seconds since epoch.*/
+    public long time; // TODO(brettmorgan): figure out how to convert this to a DateTime.
+  }
+
+  /** reviews is an array of up to five reviews. */
```
samthor: Will this always be five? Could you just say "It will contain a small number of..." (happy to defer to other libs if they do something different)

src/main/java/com/google/maps/model/PlaceDetails.java
```diff
@@ -0,0 +1,197 @@
...
+
+  /** url contains the URL of the official Google page for this place. This will be the
+   * establishment's Google+ page if the Google+ page exists, otherwise it will be the
+   * Google-owned page that contains the best available information about the place. Applications
+   * must link to or embed this page on any screen that shows detailed results about the place to
+   * the user. */
+  public URL url;
+
+  /** utcOffset contains the number of minutes this place’s current timezone is offset from UTC. */
+  public int utcOffset;
```
samthor: we don't return a 'real' timezone value?

src/main/java/com/google/maps/model/PlaceDetails.java
```diff
@@ -0,0 +1,197 @@
...
+  public String[] htmlAttributions;
+
+  /** id is deprecated. Please do not use. This field may stop being returned at any time.
+   *
+   * @deprecated This field is deprecated. See the deprecation warning at the top of
+   * <a href="https://developers.google.com/places/web-service/details">this page</a>
+   */
+  public String id;
+
+  /** reference is deprecated. Please do not use. This field may stop being returned at any time.
```
samthor: Should we just not include it then?

src/main/java/com/google/maps/model/PlaceIdScope.java
```diff
@@ -0,0 +1,15 @@
+package com.google.maps.model;
+
+/**
+ * Place ID scope is the scope of a Place ID returned from the Google Places API Web Service.
+ */
+public enum PlaceIdScope {
+  /**
+   * APP indicates the place ID is recognised by your application only.
```
samthor: Care to elaborate on this? It's a bit confusing.

src/main/java/com/google/maps/model/Photo.java
```diff
@@ -0,0 +1,28 @@
+package com.google.maps.model;
+
+/**
+ * <p>This describes a photo available with a Search Result.</p>
+ *
+ * <p>Please see <a href="https://developers.google.com/places/web-service/photos">Photos</a> for more details.</p>
+ */
+public class Photo {
```
samthor: How do I use this class to display an image? Is it missing a URL?

src/main/java/com/google/maps/model/PlacesSearchResult.java
```diff
@@ -0,0 +1,69 @@
+package com.google.maps.model;
+
+import java.net.URL;
+
+/**
+ * <p>PlaceSearchResult represents a single result in the search results return from the Google Places API Web
```
samthor: First line in Javadoc doesn't need `<p>`

src/main/java/com/google/maps/model/PlacesSearchResult.java
```diff
@@ -0,0 +1,69 @@
...
+ * Service.</p>
+ *
+ * <p>Please see <a href="https://developers.google.com/places/web-service/search#PlaceSearchResults">Place
+ * Search Results</a> for more detail.</p>
+ */
+public class PlacesSearchResult {
```
samthor: Why is this here? It seems similar but different to `PlaceDetails`.

src/main/java/com/google/maps/model/QueryAutocompletePrediction.java
```diff
@@ -0,0 +1,48 @@
...
+   * <a href="https://developers.google.com/places/web-service/query#deprecation">the deprecation notice</a>.
+   */
+  public String reference;
+
+  /**
+   * The Place ID of the place.
+   */
+  public String placeId;
+
+  public String types[];
```
samthor: needs comment

src/main/java/com/google/maps/model/QueryAutocompletePrediction.java
```diff
@@ -0,0 +1,48 @@
...
+
+  // TODO(brettmorgan): Document this on https://developers.google.com/places/web-service/query
+  public static class MatchedSubstring {
```
samthor: Is it worth making these pair classes final/with constructors?

```java
public static final class MatchedSubstring {
  public final int length;
  public final int offset;

  MatchedSubstring(...) {
    ... 
  } 
}
```

src/test/java/com/google/maps/PlacesApiIntegrationTest.java
```diff
@@ -0,0 +1,158 @@
...
+    assertEquals(placeDetails.name, "Google");
+
+    // Opening Hours
+    assertNotNull(placeDetails.openingHours);
+    assertNotNull(placeDetails.openingHours.openNow);
+    assertNotNull(placeDetails.openingHours.periods);
+    assertNotNull(placeDetails.openingHours.weekdayText);
+    assertNotNull(placeDetails.utcOffset);
+
+    // Sydney varies between GMT+10 and GMT+11
```
samthor: Technically it's either GMT+10 or GMT+11.

samthor: Also, make the comment UTC since that's the var name (and GMT ~= UTC anyway).

src/main/java/com/google/maps/PlaceDetailsRequest.java
domesticmouse: Fixed.

src/main/java/com/google/maps/PlaceDetailsRequest.java
domesticmouse: Fixed.

src/main/java/com/google/maps/PlacesApi.java
domesticmouse: From memory we have a mix across our different libraries. We have `placeID` in the iOS library, `placeid` as the URL parameter, and `place_id` in the returned JSON. In this library, my hand is forced by GSON's name mapping to use `placeId` when mapping from JSON to POJOs, so I figure I might as well be self consistent, if nothing else.

src/main/java/com/google/maps/internal/DayOfWeekAdaptor.java
domesticmouse: Agreed, much tidier.

src/main/java/com/google/maps/internal/PriceLevelAdaptor.java
domesticmouse: Done.

src/main/java/com/google/maps/model/OpeningHours.java
domesticmouse: My feeling is that we shouldn't use `java.time.DayOfWeek` as that introduces a Java 1.8 dependency. Thoughts?

samthor: That sounds fine, I wasn't sure what version of Java it was added. If it's Java 8, then definitely ignore it and use this guy.

src/main/java/com/google/maps/model/OpeningHours.java
domesticmouse: Whether these fields are set to null or not depends on the JSON we are mapping over. I don't know the edge cases of the JSON generation to know if we can make a strong claim about this field being non null.

src/main/java/com/google/maps/model/Photo.java
domesticmouse: I'm wondering how best to handle this in the API. We have another end point to add that returns the raw image itself, which means we have to prevent the GSON infrastructure kicking in. However, we still want the API key and URL Signing code to all work. Paging @markmcd for input...

src/main/java/com/google/maps/model/PlaceDetails.java
domesticmouse: Done.

src/main/java/com/google/maps/model/PlaceDetails.java
domesticmouse: Happy to be swayed either way on this one. I'm in two minds. Thoughts @markmcd?

src/main/java/com/google/maps/model/QueryAutocompletePrediction.java
domesticmouse: That'd require putting in a GSON mapper. I'm unsure of the gain here?

src/main/java/com/google/maps/model/PlaceDetails.java
domesticmouse: Sadly, no. See [PlaceDetailsResults](https://developers.google.com/places/web-service/details#PlaceDetailsResults)

src/main/java/com/google/maps/QueryAutocompleteRequest.java
domesticmouse: Done

src/main/java/com/google/maps/model/QueryAutocompletePrediction.java
samthor: sure, that's fine. I wasn't sure of how there were constructed, at this
point in reading.

Let's talk tomorrow about the possibility of using something like
ValueType, though.

src/main/java/com/google/maps/internal/PriceLevelAdaptor.java
```diff
@@ -0,0 +1,72 @@
...
+        case 2:
+          return PlaceDetails.PriceLevel.MODERATE;
+        case 3:
+          return PlaceDetails.PriceLevel.EXPENSIVE;
+        case 4:
+          return PlaceDetails.PriceLevel.VERY_EXPENSIVE;
+        default:
+          return PlaceDetails.PriceLevel.UNKNOWN;
+      }
+    } else {
```
samthor: Can you do the same single default case here? `UNKNOWN` at the end of function?

src/main/java/com/google/maps/model/OpeningHours.java
samthor: What does Python/Go do in this situation?

src/main/java/com/google/maps/internal/PriceLevelAdaptor.java
domesticmouse: Done

src/test/java/com/google/maps/PlacesApiIntegrationTest.java
domesticmouse: Done

src/main/java/com/google/maps/model/PlaceIdScope.java
domesticmouse: Done

src/main/java/com/google/maps/model/PlaceDetails.java
domesticmouse: I believe so, as this is the documentation we have up on our [API Documentation for Place Details](https://developers.google.com/places/web-service/details#PlaceDetailsResults)

src/main/java/com/google/maps/model/OpeningHours.java
domesticmouse: Converted Boolean -> boolean here, as not present in the result is the same as false.

src/main/java/com/google/maps/model/OpeningHours.java
domesticmouse: Added java doc. Turning this from Boolean -> boolean leads to collapsing Closed and Unknown states into one.

src/main/java/com/google/maps/model/OpeningHours.java
domesticmouse: Reverted this change to be consistent with above.

src/main/java/com/google/maps/PlacesApi.java
```diff
@@ -0,0 +1,108 @@
...
+ * <p>See also: <a href="https://developers.google.com/places/web-service/">Places API Web Service
+ * documentation</a>.
+ */
+public class PlacesApi {
+
+  private PlacesApi() {
+  }
+
+  /*
+   * Please note: We are specifically not implementing:
```
markmcd: TODO: remove?

src/main/java/com/google/maps/PlacesApi.java
```diff
@@ -0,0 +1,108 @@
...
+   *  - Nearby Search
+   *  - Radar Search
+   *  - Place Add
+   *  - Place Autocomplete
+   */
+
+  /**
+   * The Google Places API Text Search Service is a web service that returns information about a
```
markmcd: First sentence of javadoc should be a very short noun or verb phrase, not a full sentence. e.g. "Performs a text search for places. The Google Places API...". This is pretty important for the API call entry methods.

src/main/java/com/google/maps/PlacesApi.java
markmcd: Style guide is to treat abbreviations as words, so placeId is correct.

src/main/java/com/google/maps/model/QueryAutocompletePrediction.java
domesticmouse: Done

src/main/java/com/google/maps/PlacesApi.java
domesticmouse: I was planning on leaving this in, so people don't submit patches implementing these API endpoints. Happy to remove, however.

src/main/java/com/google/maps/model/OpeningHours.java
```diff
@@ -0,0 +1,58 @@
+package com.google.maps.model;
+
+public class OpeningHours {
+  /**
+   * openNow is a boolean value indicating if the place is open at the current time.
+   */
+  public Boolean openNow;
+
+  static public class Period {
```
markmcd: Could you reuse Joda's concept of a DayTime [Period](http://joda-time.sourceforge.net/apidocs/org/joda/time/Period.html) here?

src/main/java/com/google/maps/model/Photo.java
markmcd: Can we add a method here too? Photo.download() or similar?

src/main/java/com/google/maps/model/OpeningHours.java
domesticmouse: Although this has the same name as Joda's Period seems to be a very different concept. Joda's period is a length of time, where as this class is holding the opening and closing times of a Place over a week. Am I misunderstanding your intent?

src/main/java/com/google/maps/model/PlaceDetails.java
```diff
@@ -0,0 +1,197 @@
...
+   *
+   * @deprecated Undocumented field.
+   */
+  @Deprecated
+  public String adrAddress;
+
+  /** geometry is the location of the Place. */
+  public Geometry geometry;
+
+  /** icon contains the URL of a suggested icon which may be displayed to the user when indicating
```
markmcd: nit: there are a few comments around that are multi-line and start on the /** line, can you move them so the /** has no following text, unless it all fits on one line?

src/main/java/com/google/maps/TextSearchRequest.java
```diff
@@ -0,0 +1,152 @@
...
+    if (radius > 50000) {
+      throw new IllegalArgumentException("The maximum allowed radius is 50,000 meters.");
+    }
+    return param("radius", String.valueOf(radius));
+  }
+
+  /**
+   * minprice restricts to places that are at least this price level.
+   */
+  public TextSearchRequest minprice(int priceLevel) {
```
markmcd: Should this be using the PriceLevel enums instead of ints?

src/main/java/com/google/maps/PlacesApi.java
domesticmouse: Done

src/main/java/com/google/maps/model/PlaceDetails.java
markmcd: I'd vote to leave it out. We're in the fortunate case where we are building a new library so we can leave out the legacy we don't want to support. If it becomes a problem we can always add it in later.

Ditto for anything that's being introduced as @deprecated.

src/main/java/com/google/maps/model/PlaceIdScope.java
markmcd: I think it's covered pretty well in the docs so wouldn't worry too much. Maybe just copy the rest of the doc description across?

src/main/java/com/google/maps/model/QueryAutocompletePrediction.java
markmcd: I'm 99% sure that GSON will set final fields.

src/test/java/com/google/maps/PlacesApiIntegrationTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+    assertNotNull(placeDetails);
+
+    // Address
+    assertNotNull(placeDetails.addressComponents);
+    assertNotNull(placeDetails.formattedAddress);
+    assertEquals(placeDetails.formattedAddress, "5, 48 Pirrama Rd, Pyrmont NSW 2009, Australia");
+    assertNotNull(placeDetails.vicinity);
+    assertEquals(placeDetails.vicinity, "5 48 Pirrama Road, Pyrmont");
+    assertNotNull(placeDetails.adrAddress);
```
markmcd: adrAddress is undocumented/deprecated, why use it in the tests?

src/test/java/com/google/maps/PlacesApiIntegrationTest.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+    // Place ID
+    assertNotNull(placeDetails.placeId);
+    assertEquals(placeDetails.placeId, GOOGLE_SYDNEY);
+    assertNotNull(placeDetails.scope);
+    assertEquals(placeDetails.scope, PlaceIdScope.GOOGLE);
+    assertNotNull(placeDetails.types);
+    assertEquals(placeDetails.types[0], "establishment");
+    assertNotNull(placeDetails.rating);
+    assertEquals(placeDetails.rating, 4.4, 1.0);
```
markmcd: This test seems particularly likely to change over time, causing flakiness without adding much value.

src/test/java/com/google/maps/PlacesApiIntegrationTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+  @Test
+  public void testTextSearch() throws Exception {
+    PlacesSearchResponse response = PlacesApi.textSearchQuery(context, "Google Sydney").await();
+
+    assertNotNull(response);
+
+    assertNotNull(response.results);
+    assertEquals(1, response.results.length);
+    {
```
markmcd: What's your thinking behind all of the scoped blocks in the tests?

src/test/java/com/google/maps/PlacesApiIntegrationTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+  @Test
+  public void testPizzaInNewYork() throws Exception {
+    PlacesSearchResponse response = PlacesApi.textSearchQuery(context, "Pizza in New York").await();
+    assertNotNull(response);
+    assertNotNull(response.results);
+    assertEquals(20, response.results.length);
+    assertNotNull(response.nextPageToken);
+
+    // The returned page token is not valid for a couple of seconds.
```
markmcd: What the?!

src/test/java/com/google/maps/PlacesApiTest.java
```diff
@@ -0,0 +1,540 @@
...
+      throw new IllegalArgumentException("filename '" + filename + "' resulted in null or empty body");
+    }
+    return body;
+  }
+
+  @Test
+  public void testPlaceDetailsRequest() throws Exception {
+    MockResponse response = new MockResponse();
+    response.setBody("");
+    MockWebServer server = new MockWebServer();
```
markmcd: Move the server to the class level (@BeforeClass / @Before, whatever it is), along with the context.setBaseUrl...

src/main/java/com/google/maps/PlacesApi.java
markmcd: Fair enough. It doesn't really give any background though, is there anything we can actually say here?

src/main/java/com/google/maps/model/PlaceDetails.java
domesticmouse: Cleaned up this file's javadoc. Thanks!

src/main/java/com/google/maps/model/OpeningHours.java
markmcd: That sounds like 7 lists of Joda-Periods, each at a Joda-Instant? I was just thinking it'd be great if we could prevent reinventing the wheel (and prevent forcing developers who use this data from writing the same code over & over), but if doing so will be more complex then leave it as it is.

src/test/java/com/google/maps/PlacesApiIntegrationTest.java
domesticmouse: I was using it to scope the introduced variables, `result` in this case.

src/test/java/com/google/maps/PlacesApiIntegrationTest.java
domesticmouse: Business requirement.

src/test/java/com/google/maps/PlacesApiIntegrationTest.java
domesticmouse: That was the reason I had a +/- of 1.0 on equality testing. Deleted anyways. =)

src/test/java/com/google/maps/PlacesApiIntegrationTest.java
markmcd: Ugh, [so it is](https://groups.google.com/forum/#!topic/google-places-api/Ol8OQygDkXQ). Could you add something to the pagetoken() method and the main text search request method to explain this condition?

domesticmouse: The nextPageToken javadoc contains the following:

There is a short delay between when this response is issued, and when nextPageToken will become valid to execute.

I'll pull it out as a note on a separate line, but do you feel we need to say more here?

src/test/java/com/google/maps/PlacesApiIntegrationTest.java
domesticmouse: Deleted

src/main/java/com/google/maps/TextSearchRequest.java
domesticmouse: Good catch, fixed.

src/main/java/com/google/maps/PlacesApi.java
domesticmouse: TBA

src/test/java/com/google/maps/PlacesApiTest.java
domesticmouse: Joy of imperative APIs, `server.getPort()` (called in context of `context.seBaseUrl(...)`) cannot be called before `server.play()`.

src/main/java/com/google/maps/model/PlaceDetails.java
domesticmouse: Done

src/main/java/com/google/maps/model/Photo.java
domesticmouse: I had a go at doing something along those lines for `nextPageToken` - but it wound up involving given the result object knowledge of the context the original request was made on. 

I feel for API regularity I need to introduce a method on `PlacesApi` that takes a `photoReference` as the argument. However, I still need to modify that request such that GSON isn't used to attempt to marshall raw PNG data. Hmm.

src/main/java/com/google/maps/model/PlacesSearchResult.java
domesticmouse: As discussed IRL, yes this is bad OO design, however it exposes the returned results in an obvious way that is easy to read without chasing javadoc inheritance chains.

src/test/java/com/google/maps/PlacesApiIntegrationTest.java
markmcd: I must have missed that. My bad!

domesticmouse:
@samthor & @markmcd, I believe this is ready to submit. PTAL =)

README.md
```diff
@@ -114,6 +117,7 @@ https://developers.google.com/maps/.
  - [Distance Matrix API]
  - [Elevation API]
  - [Geocoding API]
+ - [Places API]
```
samthor: unrelated fix: should you link to [Roads API] here too?

src/main/java/com/google/maps/PhotoRequest.java
```diff
@@ -0,0 +1,96 @@
...
+   * @param maxWidth Specifies the maximum desired width, in pixels, of the image returned by the Place Photos service.
+   * @return Returns the configured PhotoRequest.
+   */
+  public PhotoRequest maxWidth(int maxWidth) {
+    return param("maxwidth", String.valueOf(maxWidth));
+  }
+
+  public static class Response implements ApiResponse<PhotoResult> {
+    // This class is here to keep the type system happy.
+    // It is not actually instantiated by OkHttpPendingResult.
```
samthor: Could you confirm this by setting a private constructor?

src/main/java/com/google/maps/PlaceDetailsRequest.java
```diff
@@ -0,0 +1,77 @@
...
+  static final ApiConfig API_CONFIG = new ApiConfig("/maps/api/place/details/json")
+      .fieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES);
+
+  public PlaceDetailsRequest(GeoApiContext context) {
+    super(context, API_CONFIG, Response.class);
+  }
+
+  /** Get the Place Details for the specified Place ID. Required. */
+  PlaceDetailsRequest placeId(String placeId) {
```
samthor: Should this be public? Users can't set this from outside this package or unless they subclass this.

src/main/java/com/google/maps/PhotoRequest.java
```diff
@@ -0,0 +1,96 @@
...
+    super(context, API_CONFIG, Response.class);
+  }
+
+  @Override
+  protected void validateRequest() {
+    if (!params().containsKey("photoreference")) {
+      throw new IllegalArgumentException("Request must contain  'photoReference'.");
+    }
+    if (!params().containsKey("maxheight") && !params().containsKey("maxwidth")) {
+      throw new IllegalArgumentException("Request must contain  'maxHeight' or 'maxWidth.");
```
samthor: You have an awkward space before 'maxHeight' here. You seem to have it in a lot of your `IllegalArgumentException` constructors. Fix?

src/main/java/com/google/maps/QueryAutocompleteRequest.java
```diff
@@ -0,0 +1,104 @@
...
+  /**
+   * radius is the distance (in meters) within which to return place results. Note that setting a
+   * radius biases results to the indicated area, but may not fully restrict results to the specified area.
+   */
+  public QueryAutocompleteRequest radius(int radius) {
+    return param("radius", String.valueOf(radius));
+  }
+
+  public static class Response implements ApiResponse<QueryAutocompletePrediction[]> {
+
```
samthor: remove blank line here, and one of the two blank lines on 84-85

src/main/java/com/google/maps/TextSearchRequest.java
```diff
@@ -0,0 +1,145 @@
...
+    if (radius > 50000) {
+      throw new IllegalArgumentException("The maximum allowed radius is 50,000 meters.");
+    }
+    return param("radius", String.valueOf(radius));
+  }
+
+  /**
+   * minprice restricts to places that are at least this price level.
+   */
+  public TextSearchRequest minprice(PlaceDetails.PriceLevel priceLevel) {
```
samthor: minPrice, maxPrice, openNow, pageToken, etc. To be consistent with what you're doing in `PhotoRequest` and other classes.

src/main/java/com/google/maps/internal/PriceLevelAdaptor.java
```diff
@@ -0,0 +1,70 @@
...
+
+  @Override
+  public PlaceDetails.PriceLevel read(JsonReader reader) throws IOException {
+    if (reader.peek() == JsonToken.NULL) {
+      reader.nextNull();
+      return null;
+    }
+
+    if (reader.peek() == JsonToken.NUMBER) {
+      int day = reader.nextInt();
```
samthor: Not a `day`, it's a `priceLevel`

src/main/java/com/google/maps/model/OpeningHours.java
```diff
@@ -0,0 +1,79 @@
...
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+ * ANY KIND, either express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.google.maps.model;
+
+import org.joda.time.LocalTime;
+
+public class OpeningHours {
```
samthor: quick JavaDoc?

src/main/java/com/google/maps/model/OpeningHours.java
```diff
@@ -0,0 +1,79 @@
...
+         * Services should be updated to support the new value.
+         */
+        UNKNOWN
+      }
+
+      /**
+       * Day that this Open/Close pair is for.
+       */
+      public Period.OpenClose.DayOfWeek day;
+      /**
```
samthor: blank line insert here

src/main/java/com/google/maps/model/OpeningHours.java
samthor: Maybe also comment on the nullability of fields here - will periods always be set, even if empty? Will all of them have open and close times?

src/main/java/com/google/maps/model/OpeningHours.java
```diff
@@ -0,0 +1,79 @@
...
+
+  /**
+   * periods is an array of opening periods covering seven days, starting from Sunday, in
+   * chronological order.
+   */
+  public Period[] periods;
+
+  /**
+   * weekdayText is an array of seven strings representing the formatted opening hours for each
+   * day of the week.
```
samthor: ", for example: "10am - 5pm"

src/main/java/com/google/maps/model/PlaceDetails.java
```diff
@@ -0,0 +1,220 @@
...
+      /** type is the name of the aspect that is being rated. */
+      public RatingType type;
+
+      /** rating is the user's rating for this particular aspect, from 0 to 3. */
+      public int rating;
+    }
+
+    /**
+     * aspects contains a collection of AspectRating objects, each of which provides a rating of
+     * a single attribute of the establishment.
```
samthor: ..."This will only be available in the Paid Maps API (whatever it's called)

src/main/java/com/google/maps/model/QueryAutocompletePrediction.java
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+ * ANY KIND, either express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.google.maps.model;
+
+public class QueryAutocompletePrediction {
```
samthor: javadoc

src/main/java/com/google/maps/model/QueryAutocompletePrediction.java
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * terms contains an array of terms identifying each section of the returned description (a section
+   * of the description is generally terminated with a comma). Each entry in the array has a value
+   * field, containing the text of the term, and an offset field, defining the start position of this
+   * term in the description, measured in Unicode characters.
+   */
+  public Term terms[];
+
+  /**
+   * MatchedSubstring describe sthe location of the entered term in the prediction result text, so
```
samthor: describe sthe => "describes the"

samthor:
Just a few last nits, looking good though.

README.md
domesticmouse: done

src/main/java/com/google/maps/PhotoRequest.java
domesticmouse: done

src/main/java/com/google/maps/TextSearchRequest.java
markmcd: There's inconsistency here in the capitalisation - minprice/maxprice where later you've used minWidth/maxWidth.

src/main/java/com/google/maps/internal/OkHttpPendingResult.java
```diff
@@ -194,6 +201,22 @@ private T parseResponse(OkHttpPendingResult<T, R> request, Response response) th
...
+    R resp;
+    String contentType = response.header("Content-Type");
+
+    // Places Photo API special case
+    if (contentType != null &&
+        contentType.startsWith("image") &&
+        responseClass == PhotoRequest.Response.class &&
+        response.code() == 200) {
+      // Photo API response is just a raw image byte array.
+      PhotoResult result = new PhotoResult();
```
markmcd: Given that images are typically larger than JSON I think we should return either (or both) a java.awt.Image object, or the raw InputStream. Thoughts @domesticmouse @samthor ?

src/main/java/com/google/maps/PlaceDetailsRequest.java
domesticmouse: Making this public. Oddly enough, the `PlacesApi.placeDetails(...)` factory method enforces that API users always supply a `placeId` anyway.

src/main/java/com/google/maps/PhotoRequest.java
domesticmouse: Fixed. Thanks!

src/main/java/com/google/maps/QueryAutocompleteRequest.java
domesticmouse: done

src/main/java/com/google/maps/internal/PriceLevelAdaptor.java
domesticmouse: done

src/main/java/com/google/maps/model/OpeningHours.java
domesticmouse: Added

src/main/java/com/google/maps/model/PlaceDetails.java
domesticmouse: done

src/main/java/com/google/maps/internal/OkHttpPendingResult.java
domesticmouse: Converting a `byte[]` into a `java.awt.Image` makes the production code dependent on `javax.image.ImageIO`, if I'm understanding StackOverflow correctly. Given the common case of caching and then serving the image, a `byte[]` and the server reported `Content-Type` serves both of those use cases. I'll add documentation pointing people at Places Library for retrieving images client side.

domesticmouse: As for raw `InputStream`, I'm unsure how to how to wedge that into `OkHttpPendingResult` such that all the retry and URL signing architecture continues to work. How about we look at that in follow up PR?

src/main/java/com/google/maps/model/PlaceDetails.java
samthor: ... and will be {@code null} otherwise
?

samthor:
LGTM!

markmcd:
LGTM

Squash when you merge.

domesticmouse:
Submitted as https://github.com/googlemaps/google-maps-services-java/pull/106

PR #114, opened 2015-11-03
xFixing the precision issue

aniruddha-nitb:
There is a loss of precision as %f will return only 6 digits after the decimal.
While the lat lng contains 7 digits after decimal. So while rounding them to 6 digits there is a loss of precision which in realtime is leading to a difference of 15 minutes.

@broady @domesticmouse please review.

### DIFF
--- a/src/main/java/com/google/maps/model/LatLng.java
+++ b/src/main/java/com/google/maps/model/LatLng.java
@@ -50,6 +50,6 @@ public class LatLng implements UrlValue {
   @Override
   public String toUrlValue() {
     // Enforce Locale to English for double to string conversion
-    return String.format(Locale.ENGLISH, "%f,%f", lat, lng);
+    return String.format(Locale.ENGLISH, "%s,%s", lat, lng);
   }
 }


domesticmouse:
@aniruddha-nitb please sign the CLA so I can land your change. Thanks!

src/main/java/com/google/maps/model/LatLng.java
```diff
@@ -50,6 +50,6 @@ public String toString() {
   @Override
   public String toUrlValue() {
     // Enforce Locale to English for double to string conversion
-    return String.format(Locale.ENGLISH, "%f,%f", lat, lng);
+    return String.format(Locale.ENGLISH, "%s,%s", lat, lng);
```
domesticmouse: I believe this will fail in certain European countries, where the decimal point is replaced with a comma in locale rendered numbers. We need the number to be in english format for the API server to grok the number. 

Can you raise an issue that demonstrates the precision issue you are having instead of just supplying a fix?

broady:
I agree that 6 digits is plenty of precision. Can just use `%9f` if we want more precision.

domesticmouse:
This issue was fixed with https://github.com/googlemaps/google-maps-services-java/pull/153

Thanks!

PR #118, opened 2015-11-26
Introducing Geocoded Waypoints into Directions API result.

domesticmouse:
I had to break API backward compatibility with this PR, as we were previously returning an array.


domesticmouse:
PTAL @markmcd

src/main/java/com/google/maps/DirectionsApi.java
```diff
@@ -47,9 +49,10 @@ public static DirectionsApiRequest getDirections(GeoApiContext context,
     return newRequest(context).origin(origin).destination(destination);
   }
 
-  static class Response implements ApiResponse<DirectionsRoute[]> {
+  static class Response implements ApiResponse<DirectionsResult> {
```
markmcd: DirectionsApi.getDirections(...) is a nice shorthand that lets a user jump straight to the data, compared with newRequest(..) that let's them tweak more things.

Can you restrict the break to just DirectionsApi.newRequest(..) instead? e.g. by introducing a FullDirectionsApiRequest subclass?

There's likely be some tradeoffs we need to make, so could warrant a quick chat IRL with the team.

src/main/java/com/google/maps/model/GeocodedWaypoint.java
```diff
@@ -0,0 +1,47 @@
...
+ * requested waypoints, or the destination. Please see
+ * <a href="https://developers.google.com/maps/documentation/directions/intro#GeocodedWaypoints">
+ *   Geocoded Waypoints</a> for more detail.
+ */
+public class GeocodedWaypoint {
+  /**
+   * {@code geocoderStatus} indicates the status code resulting from the geocoding operation. This
+   * field may contain the values {@code "OK"} or {@code "ZERO_RESULTS"}.
+   */
+  public String geocoderStatus;
```
markmcd: In the Distance Matrix API, we have these types as enums. Can you extract a common type that we can use in both places?

src/main/java/com/google/maps/model/GeocodedWaypoint.java
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * {@code placeId} is a unique identifier that can be used with other Google APIs.
+   */
+  public String placeId;
+
+  /**
+   * {@code types} indicates the address type of the geocoding result used for calculating
+   * directions.
+   */
+  public String types[];
```
markmcd: AddressType[] ?

src/main/java/com/google/maps/model/GeocodedWaypoint.java
domesticmouse: I'd prefer to introduce a new enum here, so as not to mislead developers about the potential returned values.

src/main/java/com/google/maps/model/GeocodedWaypoint.java
domesticmouse: Done

src/main/java/com/google/maps/DirectionsApi.java
domesticmouse: Introducing a full results vs an elided just the routes result was a reasonably light change. WDYT?

domesticmouse:
PTAL @markmcd =)

domesticmouse:
And @samthor

markmcd:
Looks good! It might be nice to get another pair of eyes over this for sanity's sake, but I wouldn't consider it blocking.

LGTM

domesticmouse:
Ping @samthor =)

src/main/java/com/google/maps/DirectionsApiRequest.java
```diff
@@ -15,207 +15,18 @@
...
-import com.google.maps.model.TransitRoutingPreference;
-import com.google.maps.model.TravelMode;
-import com.google.maps.model.Unit;
-
-import org.joda.time.ReadableInstant;
 
 /**
  * Request for the Directions API.
+ *
+ * <p>This request generates an elided result with just the routes.</p>
```
samthor: For the full request, use `{@link Full...}`

src/main/java/com/google/maps/model/DirectionsResult.java
```diff
@@ -0,0 +1,41 @@
...
+   * {@code geocodedWaypoints} contains an array with details about the geocoding of origin,
+   * destination and waypoints. See
+   * <a href="https://developers.google.com/maps/documentation/directions/intro#GeocodedWaypoints">
+   *   Geocoded Waypoints</a> for more detail.
+   *
+   */
+  public GeocodedWaypoint geocodedWaypoints[];
+
+  /**
+   * routes contains an array of routes from the origin to the destination. See
```
samthor: `{@code routes}` to match above

samthor: Or drop the equivalent from geocodedWaypoints. Whatever makes the most sense.

src/main/java/com/google/maps/DirectionsApiRequestBase.java
```diff
@@ -0,0 +1,220 @@
...
+import com.google.maps.model.TransitMode;
+import com.google.maps.model.TransitRoutingPreference;
+import com.google.maps.model.TravelMode;
+import com.google.maps.model.Unit;
+import org.joda.time.ReadableInstant;
+
+import static com.google.maps.internal.StringJoin.join;
+
+/**
+ * Common base class for DirectionsApiRequest and FullDirectionsApiRequest.
```
samthor: Sorry, it's really unclear to me how the different subclasses cause a different request to be sent up to the server. Can you explain it to me?

domesticmouse: It's not that the requests are different, it's that the returned data is marshalled differently. In one case we return just the route, while in the other we return route an geocoded waypoints.

src/main/java/com/google/maps/DirectionsApiRequest.java
domesticmouse: Added commentary to both DirectionsApiRequest.java and FullDirectionsApiRequest.java

src/main/java/com/google/maps/model/DirectionsResult.java
domesticmouse: I went with {@code routes}

domesticmouse:
PTAL @samthor - landed the Directions API simplification.

samthor:
LGTM

PR #128, opened 2016-02-04
NPE check for route fares, and a better test for permanently closed.

domesticmouse:
PTAL @markmcd

### DIFF
--- a/src/test/java/com/google/maps/PlacesApiIntegrationTest.java
+++ b/src/test/java/com/google/maps/PlacesApiIntegrationTest.java
@@ -18,6 +18,7 @@ package com.google.maps;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import com.google.maps.model.Photo;
 import com.google.maps.model.PhotoResult;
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/maps/PlacesApiIntegrationTest.java
```diff
@@ -206,10 +207,14 @@ public void testPlaceDetailsInFrench() throws Exception {
...
     PlacesSearchResponse response = PlacesApi.textSearchQuery(context, "ABC Learning Centres in australia").await();
     assertNotNull(response);
-    PlacesSearchResult result = response.results[0];
-    assertNotNull(result);
-    assertEquals("ABC Learning Centre", result.name);
-    assertEquals(true, result.permanentlyClosed);
+    for (PlacesSearchResult result: response.results) {
+      assertNotNull(result);
+      if(result.permanentlyClosed) {
+        // test success condition
```
markmcd: Tiny nit: you'd usually test success conditions with an assert statement, e.g. assertTrue(result.permanentlyClosed).

markmcd:
LGTM

src/test/java/com/google/maps/PlacesApiIntegrationTest.java
domesticmouse: I've had to water the test down from asserting the first returned result for the query is permanently closed, to at least one of the returned result set is permanently closed. Because data.

markmcd:
Ahh I missed the outer loop. Makes sense.

PR #135, opened 2016-02-24
Bug fixes and cleanups

domesticmouse:
PTAL @samthor && @markmcd

### DIFF
--- a/src/main/java/com/google/maps/model/AddressType.java
+++ b/src/main/java/com/google/maps/model/AddressType.java
@@ -206,9 +206,16 @@ public enum AddressType implements UrlValue {
   PLACE_OF_WORSHIP("place_of_worship"),
 
   /**
-   * {@code WARD} indicates a specific type of Japanese locality, to facilitate distinction between multiple locality components within a Japanese address.
+   * {@code WARD} indicates a specific type of Japanese locality, to facilitate distinction between
+   * multiple locality components within a Japanese address.
    */
   WARD("ward"),
+
+  /**
+   * {@code POSTAL_TOWN} indicates a grouping of geographic areas, such as locality and sublocality,
+   * used for mailing addresses in some countries.
+   */
+  POSTAL_TOWN("postal_town"),
 
   /**
    * Indicates an unknown address type returned by the server. The Java Client for Google Maps
--- a/src/main/java/com/google/maps/model/PlacesSearchResult.java
+++ b/src/main/java/com/google/maps/model/PlacesSearchResult.java
@@ -67,7 +67,7 @@ public class PlacesSearchResult {
   /**
    * types contains an array of feature types describing the given result.
    */
-  public String types[];
+  public AddressType types[];
 
   /**
    * openingHours may contain whether the place is open now or not.
--- a/src/main/java/com/google/maps/model/PlacesSearchResult.java
+++ b/src/main/java/com/google/maps/model/PlacesSearchResult.java
@@ -67,7 +67,7 @@ public class PlacesSearchResult {
   /**
    * types contains an array of feature types describing the given result.
    */
-  public AddressType types[];
+  public String types[];
 
   /**
    * openingHours may contain whether the place is open now or not.
--- a/src/main/java/com/google/maps/model/Vehicle.java
+++ b/src/main/java/com/google/maps/model/Vehicle.java
@@ -41,7 +41,7 @@ public class Vehicle {
   public String icon;
 
   /**
-   * {@code local_icon} contains the URL for an icon based on the local transport signage.
+   * {@code localIcon} contains the URL for an icon based on the local transport signage.
    */
-  public String local_icon;
+  public String localIcon;
 }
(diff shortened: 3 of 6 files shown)


src/test/java/com/google/maps/DirectionsApiTest.java
```diff
@@ -394,4 +394,29 @@ public void testGeocodedWaypoints() throws Exception {
...
+   * Test {@code local_icon} for Directions in Paris.
+   */
+  @Test
+  public void testLocalIconInParis() throws Exception {
+    DirectionsResult result = DirectionsApi.newRequest(context)
+        .origin("paris metro bibliotheque francois mitterrand")
+        .destination("paris metro pyramides")
+        .mode(TravelMode.TRANSIT)
+        .await();
+    assertNotNull(result);
```
samthor: I know there's not, but I'm really sad there's not a better way to do this.

Well.. actually. You could just...

```java
String localIcon = "";
try {
  localIcon = result.routes[0].legs[0].....
} catch (NullPointerException e) {
  fail("expected transit...");
}
assertEquals(...);
```

Just my 2c, up to you though.

src/main/java/com/google/maps/model/Vehicle.java
```diff
@@ -39,4 +39,9 @@
    * {@code icon} contains the URL for an icon associated with this vehicle type.
    */
   public String icon;
+
+  /**
+   * {@code local_icon} contains the URL for an icon based on the local transport signage.
+   */
+  public String local_icon;
```
samthor: Shouldn't this be `localIcon`?

src/main/java/com/google/maps/model/PlacesSearchResult.java
```diff
@@ -67,7 +67,7 @@
   /**
    * types contains an array of feature types describing the given result.
    */
-  public String types[];
+  public AddressType types[];
```
samthor: It doesn't look like the code which sets up this value has changed? Or does the JSON decoder intelligently do the right thing for string enums here?

src/main/java/com/google/maps/model/AddressType.java
```diff
@@ -206,6 +206,18 @@
   PLACE_OF_WORSHIP("place_of_worship"),
 
   /**
+   * {@code WARD} indicates a specific type of Japanese locality, to facilitate distinction between
+   * multiple locality components within a Japanese address.
+   */
+  WARD("ward"),
```
samthor: 🇯🇵!

src/test/java/com/google/maps/DirectionsApiTest.java
domesticmouse: The long hand approach makes debugging easier. =)

src/main/java/com/google/maps/model/PlacesSearchResult.java
domesticmouse: The magic responsible for this is https://github.com/googlemaps/google-maps-services-java/blob/master/src/main/java/com/google/maps/internal/SafeEnumAdapter.java#L64

However, I'm having to revert this change because types on place search results aren't AddressTypes. They just look like them sometimes.

src/main/java/com/google/maps/model/AddressType.java
domesticmouse: That's less than understandable on my linux desktop. Yay Emoji. =)

samthor:
LGTM

src/test/java/com/google/maps/PlacesApiIntegrationTest.java
```diff
@@ -273,7 +275,7 @@ public void testRadarSearchRequestByKeyword() throws Exception {
         .keyword("pub").await();
     assertNotNull(response);
     assertNotNull(response.results);
-    assertTrue(175 < response.results.length);
+    assertTrue(100 < response.results.length);
```
markmcd: What happened to all of the pubs in Sydney? :fearful:

src/test/java/com/google/maps/DirectionsApiTest.java
markmcd: I wouldn't even catch the NPE. JUnit will catch and report the error.

For debugging purposes would it be cleaner to just log the request/response? Or use a breakpoint?

src/test/java/com/google/maps/PlacesApiIntegrationTest.java
```diff
@@ -176,8 +178,8 @@ public void testPhoto() throws Exception {
     assertNotNull(image);
   }
 
-  @Test
-  public void testPizzaInNewYork() throws Exception {
+    @Test
```
markmcd: nit: spaces

domesticmouse:
PTAL @broady

src/test/java/com/google/maps/PlacesApiIntegrationTest.java
domesticmouse: Late night trading restrictions?

src/test/java/com/google/maps/PlacesApiIntegrationTest.java
domesticmouse: Fixed

src/test/java/com/google/maps/DirectionsApiTest.java
domesticmouse: Dropping most of the assertNotNull tests.

src/main/java/com/google/maps/model/Vehicle.java
domesticmouse: Fixed

PR #139, opened 2016-03-04
distance matrix request - add traffic_model to request and durationInTraffic to response

ctmckenna:
google returns durationInTraffic in response to distance matrix requests, so adding that to DistanceMatrixElement.  Also adding ability to specify traffic model for distance matrix requests

### DIFF
--- a/src/main/java/com/google/maps/DistanceMatrixApiRequest.java
+++ b/src/main/java/com/google/maps/DistanceMatrixApiRequest.java
@@ -19,7 +19,13 @@ import static com.google.maps.internal.StringJoin.join;
 
 import com.google.maps.DirectionsApi.RouteRestriction;
 import com.google.maps.DistanceMatrixApi.Response;
-import com.google.maps.model.*;
+import com.google.maps.model.DistanceMatrix;
+import com.google.maps.model.LatLng;
+import com.google.maps.model.TransitMode;
+import com.google.maps.model.TransitRoutingPreference;
+import com.google.maps.model.TravelMode;
+import com.google.maps.model.Unit;
+import com.google.maps.model.TrafficModel;
 
 import org.joda.time.ReadableInstant;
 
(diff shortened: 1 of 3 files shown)


ctmckenna:
I signed it!

src/main/java/com/google/maps/DistanceMatrixApiRequest.java
```diff
@@ -19,12 +19,7 @@
 
 import com.google.maps.DirectionsApi.RouteRestriction;
 import com.google.maps.DistanceMatrixApi.Response;
-import com.google.maps.model.DistanceMatrix;
-import com.google.maps.model.LatLng;
-import com.google.maps.model.TransitMode;
-import com.google.maps.model.TransitRoutingPreference;
-import com.google.maps.model.TravelMode;
-import com.google.maps.model.Unit;
+import com.google.maps.model.*;
```
domesticmouse: Can you please revert this change? We don't used star imports, per the Google Java style guide lines.

src/main/java/com/google/maps/model/DistanceMatrixElement.java
```diff
@@ -35,6 +35,8 @@
    */
   public Duration duration;
 
+  public Duration durationInTraffic;
```
domesticmouse: Please add Javadoc

ctmckenna:
removed import *, and added javadoc for durationInTraffic field

domesticmouse:
Can you please add a test that confirms that the `traffic_model` parameter is included in the outbound request, and the `durationInTraffic` is populated on the way back in. Thanks!

ctmckenna:
I just pushed a test case.  It verifies that traffic_model was included in the outbound request by checking the params table.  I also manually checked that it was included in the get request.

domesticmouse:
Thank you!

PR #140, opened 2016-03-10
Fix places autocomplete types parameter and introduced correct type enum

mikkelspring:
Based on my understanding, the autocomplete for places is slightly different than the places search. In particular, it has a parameter name "types" that takes a restricted set of values.

https://developers.google.com/places/web-service/autocomplete

This PR adjusts the autocomplete to reflect this


mikkelspring:
I signed the CLA

domesticmouse:
I suspect this is a documentation bug, please hold this until I confirm one way or the other.

PR #152, opened 2016-04-28
Adding canonical literals for AddressComponentType

manjeshnilange:
I noticed that while AddressType had the canonical literals (the exact strings in the server response), AddressComponentType didn't have them.

### DIFF
--- a/src/main/java/com/google/maps/model/AddressType.java
+++ b/src/main/java/com/google/maps/model/AddressType.java
@@ -234,6 +234,10 @@ public enum AddressType implements UrlValue {
     return addressType;
   }
 
+  public String toCanonicalLiteral() {
+    return toString();
+  }
+
   @Override
   public String toUrlValue() {
     if (this == UNKNOWN) {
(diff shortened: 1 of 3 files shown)


manjeshnilange:
I signed it!

src/main/java/com/google/maps/model/AddressType.java
```diff
@@ -234,6 +234,10 @@ public String toString() {
     return addressType;
   }
 
+  public String toCanonicalLiteral() {
```
domesticmouse: I'm curious, who calls `toCanonicalLiteral`?

manjeshnilange: I'd like to. I think it's more explicit that this literal resembles the literal returned by the server. `toString` doesn't seem to promise that. And `toUrlValue` seems to indicate that this can be used in a URL query string, but not that it's the literal returned by the server.

domesticmouse: That's a reasonable request, now that I understand the context. Can you please add tests to cover the two `toCanonicalLiteral` calls you have added, just to make sure we don't break these while moving forwward. Thanks!

domesticmouse:
Thanks!

manjeshnilange:
Sure, no problem. @domesticmouse, when do you think you'll release the next version :)

domesticmouse:
@manjeshnilange I have a new feature that I'm working on that will be finished in the next couple of weeks. So, that's the time frame =)

PR #153, opened 2016-04-29
Increasing the precision of LatLng's toUrlValue

domesticmouse:
See
https://github.com/googlemaps/google-maps-services-java/pull/114#issuecomment-215319159
for context.

### DIFF
--- a/src/main/java/com/google/maps/model/LatLng.java
+++ b/src/main/java/com/google/maps/model/LatLng.java
@@ -50,6 +50,6 @@ public class LatLng implements UrlValue {
   @Override
   public String toUrlValue() {
     // Enforce Locale to English for double to string conversion
-    return String.format(Locale.ENGLISH, "%9f,%9f", lat, lng);
+    return String.format(Locale.ENGLISH, "%.8f,%.8f", lat, lng);
   }
 }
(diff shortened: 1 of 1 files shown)


broady:
Is 9 too much? Can you calculate the average distance where this would make a difference?

domesticmouse:
Reading through the original report, it looks like `%.6f` is the current state of play, which the reporter says is affecting their usage of the API. I honestly don't know what the real world distance of round seven significant digits back to six digits is, and I'm not sure who to ask.

broady:
http://www2.usgs.gov/faq/categories/9794/3022

> At 38 degrees North latitude, one degree of latitude equals approximately 364,000 ft (69 miles), one minute equals 6068 ft (1.15 miles), one-second equals 101 ft; one-degree of longitude equals 288,200 ft (54.6 miles), one minute equals 4800 ft (0.91 mile), and one second equals 80 ft.

69 miles is 111,045m

0.000001 * 111045 = 0.111045

That is, 110mm

%.6f = 110mm
%.7f = 11mm
%.8f = 1mm

That seems good enough for anything we're doing, but yeah, perhaps bumping it up is better. Anywhere between 7 and 9 seems fine.

broady:
Also, you want a period before the 9.
`%.9f`, not `%9f`

domesticmouse:
Thanks for the clarification @broady 

I think 1mm precision is good enough.

broady:
lg

PR #165, opened 2016-06-22
Adding test that shows how the library gets stuck when multiple requests are done

Ismamai:
In order to test this you need on development key, we have to make requests to get the DAILY LIMIT.

This has also happen with external keys that have bigger limits, in this case it must be OVER_QUERY_LIMIT (even the qps is set to the right value). It seems that when there is a OVER_QUERY_LIMIT( DailyLimit or really OverQueryLimit) the library tries to send a new request with okHttp, that request is never send or we never get any result (success or failure) and the library hangs forever

The test tries to force you to get over the daily limit. There are two variables to configure how many requests are sent. Initially they are configure to create 2500 requests to geocode Sydney. Once we get the limit, we can lower the number of requests. The test will never finished as the library gets stuck.

### DIFF
--- a/src/test/java/com/google/maps/MultipleGeocodingTest.java
+++ b/src/test/java/com/google/maps/MultipleGeocodingTest.java
@@ -1,0 +1,90 @@
+package com.google.maps;
+
+import com.google.maps.model.GeocodingResult;
+import com.google.maps.model.LocationType;
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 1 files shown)


domesticmouse:
When I bring this down to 250 requests, 10 per batch, 1 second between batches your test passes fine. I'm not going to accept this patch as is, because it will quickly chew up quota of anyone running the test suite.

Ismamai:
This test is to show that the library gets stuck if it there was an error calling google backend and the library tries to retry the request. In that case, the library gets stuck. This test was not thought be added as a regular test, it was mainly to show the problem.

- In development, I was able to reproduce it just by using all the daily allowance. The library just got stuck. My program did not get any result from the library and got stuck waiting for a response that never comes.

- In production using paid keys, sometimes the library gets stuck and the library never recovers. But I do not know how to reproduce this as it depends on google response causing a retry.

PR #172, opened 2016-08-24
Made API results serializable for caching

boriszamoruev:
For performance reasons we need to store Google Directions and Distance Matrix results in memcached. Having results implement Serializable makes it easier to work with Java memcached library.

### DIFF
--- a/src/main/java/com/google/maps/model/Bounds.java
+++ b/src/main/java/com/google/maps/model/Bounds.java
@@ -15,10 +15,12 @@
 
 package com.google.maps.model;
 
+import java.io.Serializable;
+
 /**
  * The north east and south west points that delineate the outer bounds of a map.
  */
-public class Bounds {
+public class Bounds implements Serializable {
   public LatLng northeast;
   public LatLng southwest;
 }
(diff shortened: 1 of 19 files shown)


domesticmouse:
At the very least you should implement `serialVersionUID` along with declaring the classes `serializable`. However, I also feel that Java's serialization has a lot of known downsides, and you'd be a lot better off using protobufs or roundtripping the underlying JSON that putting Java serialized instances into memcache.

boriszamoruev:
Could you please elaborate on "roundtripping the underlying JSON" idea. Lack of default constructors makes it difficult to deserialize Directions Results, etc. from JSON. I would really prefer to cache the original JSON returned from Google APIs but it doesn't look like there's an easy way to get it.

domesticmouse:
Yeah, it would require some open heart surgery to insert a memcache provider into the middle of the response stack. If you want to have a go at defining a service provider interface I'll work it over with you. That said, I still think storing the most minimal version of your required data as a binary protobufs in memcache will give you the best bang for your memcache dollars.

PR #174, opened 2016-08-26
Set RateLimitExecutor thread name.

Ryszard-Trojnacki:
RateLimitExecutorService thread doesn't has a name (oversight?). 
It is not a bug, but it is easer to read stacktraces with thread with names.


Ryszard-Trojnacki:
I signed it!

domesticmouse:
Thanks!

PR #177, opened 2016-09-05
Add support for the 'permanently_closed' attribute in Place responses.

amyboyd:
This attribute is only included in the Places response if the place has shut down. It is not included if the place is still open.


src/test/java/com/google/maps/PlacesApiTest.java
```diff
@@ -281,6 +281,9 @@ public void testPlaceDetailsLookupGoogleSydney() throws Exception {
     assertEquals(placeDetails.types[0], "establishment");
     assertNotNull(placeDetails.rating);
     assertEquals(placeDetails.rating, 4.4, 0.1);
+
+    // Permanently closed:
+    assertFalse(placeDetails.permanentlyClosed);
```
domesticmouse: Can you please add a test that has permanentlyClosed set to true?

amyboyd: @domesticmouse Sure - done and PR updated!

domesticmouse:
Thanks! I'll roll a new release in the next couple of weeks.

PR #181, opened 2016-09-23
Fix poor resource management code (fixes #179)

### LINKED ISSUE #179: Connection leak due to not closing input stream on error
> WARN  [2016-09-14 22:31:24,510] com.squareup.okhttp.OkHttpClient: A connection to https://maps.googleapis.com/ was leaked. Did you forget to close a response body?

`OkHttpPendingResult#getBytes()` does not close input stream. While OkHttp can recover on a 2xx, it will not do this for you on an error code. As [explained](https://github.com/square/okhttp/issues/2311), they expect callers to write proper resource management code.

When a failure occurs this library typically hangs, as seen in #161, and requires callers perform their own timeouts. Unfortunately calling `PendingResult#cancel()` afterwards is not good enough to recover, because the connection is stuck by the open input stream. Eventually the pool will be exhausted and no new requests are performed. The only resolution is to restart the application.

ben-manes:
When the response contents are copied the input stream was not being
closed. This might appear correct for a one-off tests due to new
connections being established. Unfortunately it eventually consumes
the entire pool with stuck connections, resulting in no new requests
being sent. Cancelling the request isn't enough since the connection
is still stuck in a bad state.

The simple solution is to close the InputStream, but since OkHttp has
a method to do this we'll just delete this bad code entirely.

### DIFF
--- a/src/main/java/com/google/maps/internal/OkHttpPendingResult.java
+++ b/src/main/java/com/google/maps/internal/OkHttpPendingResult.java
@@ -46,9 +46,7 @@ import org.joda.time.DateTime;
 import org.joda.time.Instant;
 import org.joda.time.LocalTime;
 
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.io.InputStream;
 import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.ArrayBlockingQueue;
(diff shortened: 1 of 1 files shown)


ben-manes:
This is available on jitpack.io using the coordinates
`com.github.ben-manes:google-maps-services-java:cc3530ad22`

PR #182, opened 2016-09-27
Test library stuck

Ismamai:
I just updated the test with some changes and rebase it on last changes made on the library that deal with resource management. Doing those changes closed my previous pull request. (https://github.com/googlemaps/google-maps-services-java/pull/165)

Those changes did not help with this test. it keeps getting stuck even with last changes.

Once you get an error the library tries to retry the request and it never comes back.

Not sure what is the expected behaviour of the methods that set timeouts on the GeoApiContext

In my test I set these values to 1 second. But it does not make any difference. I would expect an error return if these timeouts expire.

```
                .setConnectTimeout(1, TimeUnit.SECONDS)
                .setReadTimeout(1, TimeUnit.SECONDS)
                .setWriteTimeout(1, TimeUnit.SECONDS);
```

Sep 27, 2016 10:55:22 AM com.google.maps.OkHttpRequestHandler handle
INFO: Request: https://maps.googleapis.com/maps/api/geocode/json?key=API_KEY&address=Sydney
Sep 27, 2016 10:55:22 AM com.google.maps.OkHttpRequestHandler handle
INFO: Request: https://maps.googleapis.com/maps/api/geocode/json?key=API_KEY&address=Sydney
Sep 27, 2016 10:55:22 AM com.google.maps.OkHttpRequestHandler handle
INFO: Request: https://maps.googleapis.com/maps/api/geocode/json?key=API_KEY&address=Sydney
Sep 27, 2016 10:55:22 AM com.google.maps.OkHttpRequestHandler handle
INFO: Request: https://maps.googleapis.com/maps/api/geocode/json?key=API_KEY&address=Sydney
Sep 27, 2016 10:55:23 AM com.google.maps.internal.OkHttpPendingResult retry
INFO: Retrying request. Retry #1

Test will fail after the timeout for running the test expires

`Thread.sleep(10000);`

### DIFF
--- a/src/test/java/com/google/maps/MultipleGeocodingTest.java
+++ b/src/test/java/com/google/maps/MultipleGeocodingTest.java
@@ -1,0 +1,90 @@
+package com.google.maps;
+
+import com.google.maps.model.GeocodingResult;
+import com.google.maps.model.LocationType;
+import org.junit.Test;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Created by iblesa on 21/06/2016.
+ */
+public class MultipleGeocodingTest extends AuthenticatedTest {
+
+    private static final double EPSILON = 0.000001;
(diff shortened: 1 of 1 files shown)


domesticmouse:
I suspect this is no longer an issue, now that the tests aren't integration tests. Please feel free to re-open if this isn't the case.

PR #185, opened 2016-10-01
Allow retries to be limited by number of retries, not time.

amyboyd:
In some cases, it can be useful to restrict the number of retries by retry count, not just time. This pull request adds `GeoApiContext.setMaxRetries(Integer)` and `GeoApiContext.disableRetries()`.

This does not remove the ability to limit retries by time.


domesticmouse:
Thanks Amy!

PR #198, opened 2016-11-07
Changing internal ExecutorService to use newWorkStealingPool

Ismamai:
This one will grow and shrink based on the number of operations. By default it builds as many threads as current available cores. This is introduced in java 8.

If we do not want to increase library to use java 8, we could use ThreadPoolExecutor initialized to the number of current available cores.

Another issue is that it does not take a ThreadFactory as argument and we lose the possibility to name the threads and set them as daemons.


src/main/java/com/google/maps/internal/RateLimitExecutorService.java
```diff
@@ -44,9 +43,7 @@
   // It's important we set Ok's second arg to threadFactory(.., true) to ensure the threads are
```
domesticmouse: This comment now looks wrong.

Ismamai:
Comment removed :)

domesticmouse:
I went with #199 instead of this patch. Thanks!

PR #205, opened 2016-11-21
Add waypoints method that takes in an array of LatLng

amc6:
The origin and destination methods have two versions, one that takes a String and one that takes a LatLng. This commit adds a second method for the waypoints method that takes in a LatLng array.


domesticmouse:
Can you please add tests for this? Thanks!

domesticmouse:
Thanks!

PR #222, opened 2017-01-10
Correct usage of GeoApiContext

jmarcelo:
Documents correct usage of GeoApiContext, since the wrong usage could cause a lot of threads alive in JVM.


domesticmouse:
Thanks!

Issue #175, opened 2016-08-27
Create specific exception instead of generic Exception


lsoares:
```
try {
    req.await();
    // "other code"
} catch (Exception e) {
}
```

[Dealing with generic exceptions creates some dangers](http://www.javaworld.com/article/2073800/testing-debugging/beware-the-dangers-of-generic-exceptions.html). I propose that you create a specific exception (or more) like `GeocodingException` or likewise.

Otherwise, I may be be preventing runtime exceptions (and also catched exceptions) under "other code" to bubble up...

(I can also do it as PR if you prefer.)

domesticmouse:
I'm always up for reviewing PRs. Thanks!

pieteradejong:
the code fragment refers to the readme.
is the intention to update the readme, 
or to modify (make more specific) (some of) the 9 occurrences thoughout the code?

lsoares:
I meant in the code.
I tried to do the PR but it bears lots of changes..


Em 18/01/2017 04:36, "Pieter Arthur de Jong" <notifications@github.com>
escreveu:

the code fragment refers to the readme.
is the intention to update the readme,
or to modify (make more specific) (some of) the 9 occurrences thoughout the
code?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/googlemaps/google-maps-services-java/issues/175#issuecomment-273380798>,
or mute the thread

pieteradejong:
@lsoares do you see a path toward breaking it up into smaller pieces?

lsoares:
What if `await` under `PendingResult` throws an application exception (e.g. `GoogleMapsException`) instead of `Exception`? That would be a start I guess.

domesticmouse:
I'm going to close this out for now. Please feel free to submit either the uber-PR or a series of smaller PRs. I'm about to break the world by migrating from OkHttp2 to OkHttp3, however...

PR #238, opened 2017-03-11
Make PendingResult#await throw specific Exception

daisy1754:
This PR addresses #175. Currently PendingResult#await() throws generic Exception. This PR refactor so it throws IOException (exception during HTTP communication), InterruptedException (interrupted while waiting response), ApiException (error in API response).


domesticmouse:
Thank you for this cleanup!

Issue #228, opened 2017-02-09
Question about QPS using google-maps-service


irineuruiz:
We have a premium account, so we assume we can made 50 qps for geocoding when we send various geocoding request.

We made a simple code to test our performance.

``

	private static int TOTAL_REQ = 50;
	
	private static final GeoApiContext context = new GeoApiContext().setQueryRateLimit(50, 0)
			.setApiKey("OUR_API_KEY")
			.setEnterpriseCredentials("CREDENTIAL", "CRENDENTIAL");

	public static void main(String[] args) throws Exception {
		List<LatLng> latLngs = new ArrayList<LatLng>();
		// Random locations
		Random randomGenerator = new Random();
		for (int i = 0; i < TOTAL_REQ; i++) {
			LatLng point = new LatLng(randomGenerator.nextDouble() - 23.5798154,
					randomGenerator.nextDouble() - 46.6945082);
			latLngs.add(point);
		}

		//First take always too long
		LatLng point = new LatLng(randomGenerator.nextDouble() - 23.5798154, randomGenerator.nextDouble() - 46.6945082);
		new GeocodingApiRequest(context).language("pt-BR").latlng(point).await();

		//Stopwatch clock = Stopwatch.createStarted();
		long millis = System.currentTimeMillis();
		
		final AtomicInteger counter = new AtomicInteger(TOTAL_REQ);
		for (final LatLng latLng : latLngs) {
			GeocodingApiRequest request = new GeocodingApiRequest(context).language("pt-BR");
			request.latlng(latLng).setCallback(new Callback<GeocodingResult[]>() {
				public void onResult(GeocodingResult[] result) {
					System.out.println(result[0].formattedAddress);
					counter.decrementAndGet();
				}

				public void onFailure(Throwable e) {
					e.printStackTrace();
				}
			});
		}

		while (counter.get() > 0){
			Thread.sleep(100);
		}
		millis = System.currentTimeMillis() - millis;
		System.out.println("Time taken: " + millis);
		System.out.println("QPS achieved: " + TOTAL_REQ/(double)millis/1000);
	}

``

That is the result obtained:

Time taken: 13108
QPS achieved: 3.8144644491913335E-6

Our question is: 
Why we have such QPS? Is this normal?
We expected something near 50 QPS..

We looked the network during the execution and we saw that only 2 or 3 http connections was opened by OkHttpClient to make the requests...
Looks like this is related with this lower qps.

nbluis:
[...]

PR #258, opened 2017-04-10
Solving performance problems in parallel processing

### LINKED ISSUE #256: OkHttp limiting concurrent requests
I'm using the API to perform parallel coordinate processing.

But even setting setQueriesPerSecond(20) when building GeoApiContext, apparently it performs on average only 5 querys per second, regardless of the value set.

I use an account with enterprise credentials and outside GAE.

I found the following snippets that I think might be causing the problem.

When creating the [RequestHandler](https://github.com/googlemaps/google-maps-services-java/blob/master/src/main/java/com/google/maps/OkHttpRequestHandler.java#L48) is instantiated an okhttp Dispatcher.

By default the Dispatcher has a [request limit per host](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Dispatcher.java#L40).

Since all requests to the API are made to the same host, I believe that when creating the Dispatcher it should invoke [Dispatcher.setMaxRequestsPerHost](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Dispatcher.java#L98) sending the limit coming from GeoApiContext, otherwise this setting will not work properly.

What do you think ?

I can try to verify this and send a pull request with the fix.

nbluis:
Hi.

In this patch, I first adjusted the problem described in issue #256.

But even after this, I did some testing and the parallel performance was still much lower than expected (set in GeoApiContextProvider.setQueryRateLimit)

After a few more tests, I identified that the problem was in the delay between [take operations in RateLimitExecutorService](https://github.com/googlemaps/google-maps-services-java/blob/master/src/main/java/com/google/maps/internal/RateLimitExecutorService.java#L90).

The [ThreadPoolExecutor documentation](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadPoolExecutor.html), explains the overhead of using the LinkedBlockingQueue as workQueue and suggests the use of SynchronousQueue as a good choice.

To my surprise, when I tested only this change I had incredible results.

I used the example cited by @irineuruiz in #228 with litle changes to perform the test with 500 tries and 50 QPS.

Using LinkedBlockingQueue:
Time taken: 22248
QPS achieved: 22.473930240920534

Using SynchronousQueue:
Time taken: 11467
QPS achieved: 43.60338362256911

Also, I checked that by default the [OkHttp Dispatcher](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Dispatcher.java#L65) uses SynchronousQueue to run.  So I think it's a pretty safe change.

This patch fixes #256 and #228 issues and will allows us to achiev the expected QPS.

### DIFF
--- a/src/main/java/com/google/maps/OkHttpRequestHandler.java
+++ b/src/main/java/com/google/maps/OkHttpRequestHandler.java
@@ -42,10 +42,12 @@ public class OkHttpRequestHandler implements GeoApiContext.RequestHandler {
   private static final MediaType JSON = MediaType.parse("application/json; charset=utf-8");
   private final OkHttpClient client = new OkHttpClient();
   private final RateLimitExecutorService rateLimitExecutorService;
+  private final Dispatcher dispatcher;
 
   public OkHttpRequestHandler() {
     rateLimitExecutorService = new RateLimitExecutorService();
-    client.setDispatcher(new Dispatcher(rateLimitExecutorService));
+    dispatcher = new Dispatcher(rateLimitExecutorService);
+    client.setDispatcher(dispatcher);
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


domesticmouse:
It looks like one of your commits is missing an author. Do you want to create a new PR with correct authorship details?

nbluis:
Ohh ok, will do it.

PR #260, opened 2017-04-17
Fix for issues 218,170 - aded instance creator for EncodedPolyline to use with Gson

ArielY15:


ArielY15:
I signed it!

domesticmouse:
Thanks @ArielY15!

Issue #261, opened 2017-04-20
GeoApiContext should clean up working threads on exit.


pooi-pooi:
Hello guys, I had a problem while doing a **PlacesApi.placeAutocomplete** request
I really don't know if that is a issue caused by the API or on our implementation.

We are developing an application that uses:
* Spring Boot `1.3.6.RELEASE`
* `Undertow`
* `com.google.maps:google-maps-services:0.1.17`

```
# application.properties
server.undertow.worker-threads=1000
server.undertow.io-threads=100
```

```
# EC2 AWS
m4.xlarge
```

```Dockerfile
# Dockerfile

FROM java:8
RUN apt-get update && apt-get install -y

WORKDIR /app
VOLUME /root/.gradle

COPY gradle/ /app/gradle/
COPY gradlew /app/gradlew
COPY build.gradle /app/build.gradle
COPY src/ /app/src/

COPY docker-entrypoint.sh /
ENTRYPOINT ["/docker-entrypoint.sh"]

CMD ["java","-jar","/app/build/libs/application-1.0.0.jar"]
```

```sh
# docker-entrypoint.sh

#!/bin/bash
set -e
./gradlew build -x test
exec "$@"
```

```java
@Service
public class GoogleMapsAPIService {
    public List placeAutoComplete(String input, Double latitude, Double longitude, String language) {
        GeoApiContext context = new GeoApiContext().setApiKey(GOOGLEMAPS_KEY);
        LatLng location = new LatLng(latitude, longitude); // -23.52488881961586,-46.67578987777233

        AutocompletePrediction[] autocompletePredictions = PlacesApi.placeAutocomplete(context, input)
            .location(location)
            .language("pt-BR")
            .radius(1000)
            .awaitIgnoreError();

        return assemblyPrediction(autocompletePredictions);
    }

    private List assemblyPrediction(AutocompletePrediction[] autoCompletePredictions) {
        List<Prediction> predictions = new ArrayList<>();

        for (AutocompletePrediction autocompletePrediction : autoCompletePredictions) {
            String placeId = autocompletePrediction.placeId;
            String title = autocompletePrediction.terms[0].value;
            String subTitle = Arrays.stream(autocompletePrediction.terms).skip(1).map(term -> term.value).collect(Collectors.joining(", "));

            predictions.add(new Prediction(placeId, title, subTitle));
        }

        return predictions;
    }
}
```

## **Sometimes we get this exception:**

```
[...]

PR #272, opened 2017-06-16
Replacing hand rolled rate limiter with Gauva's Rate Limiter.

domesticmouse:
PTAL @markmcd 

This is a fix for #261

### DIFF
--- a/src/main/java/com/google/maps/OkHttpRequestHandler.java
+++ b/src/main/java/com/google/maps/OkHttpRequestHandler.java
@@ -102,11 +102,9 @@ public class OkHttpRequestHandler implements GeoApiContext.RequestHandler {
     rateLimitExecutorService.setQueriesPerSecond(maxQps);
   }
 
-  @Override
+  @Override @Deprecated
   public void setQueriesPerSecond(int maxQps, int minimumInterval) {
-    dispatcher.setMaxRequests(maxQps);
-    dispatcher.setMaxRequestsPerHost(maxQps);
-    rateLimitExecutorService.setQueriesPerSecond(maxQps, minimumInterval);
+    this.setQueriesPerSecond(maxQps);
   }
 
   @Override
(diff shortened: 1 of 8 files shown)


src/main/java/com/google/maps/OkHttpRequestHandler.java
```diff
@@ -102,11 +102,9 @@ public void setQueriesPerSecond(int maxQps) {
     rateLimitExecutorService.setQueriesPerSecond(maxQps);
   }
 
-  @Override
+  @Override @Deprecated
   public void setQueriesPerSecond(int maxQps, int minimumInterval) {
-    dispatcher.setMaxRequests(maxQps);
-    dispatcher.setMaxRequestsPerHost(maxQps);
-    rateLimitExecutorService.setQueriesPerSecond(maxQps, minimumInterval);
+    this.setQueriesPerSecond(maxQps);
```
markmcd: Can you log a warning or something here? I'm concerned that ignoring an explicitly provided parameter will lead to unexpected behaviour for users.

src/test/java/com/google/maps/model/EnumsTest.java
```diff
@@ -147,6 +147,7 @@ public void testCanonicalLiteralsForAddressComponentType() {
     addressComponentTypeToLiteralMap.put(AddressComponentType.TRAIN_STATION, "train_station");
     addressComponentTypeToLiteralMap.put(AddressComponentType.SUBWAY_STATION, "subway_station");
     addressComponentTypeToLiteralMap.put(AddressComponentType.TRANSIT_STATION, "transit_station");
+    addressComponentTypeToLiteralMap.put(AddressComponentType.LIGHT_RAIL_STATION, "light_rail_station");
```
markmcd: This appears to be unrelated to your proposed change.

domesticmouse: I can put it into a separate PR, it's a fix for a test failure.

src/main/java/com/google/maps/OkHttpRequestHandler.java
domesticmouse: Done

PR #290, opened 2017-07-14
Fix issue where Runnable runs on the incorrect thread

robmarsh:
Fixes issue where Runnable was running on the RateLimitExecutorDelayThread thread, allowing only one request per rate limit to go through.   This moves it back onto the delegate ExecutorService.  

This looks like it was introduced in Commit 4bf88c29a14214e505765a35bb27ab825472bbaa.


robmarsh:
I signed it.

domesticmouse:
Thanks!

PR #293, opened 2017-07-24
Landing Okhttp3

domesticmouse:


src/main/java/com/google/maps/GeoApiContext.java
```diff
@@ -264,140 +273,226 @@ private void checkContext(boolean canUseClientId) {
...
-  /**
-   * Sets the proxy for new connections.
-   *
-   * @param proxy The proxy to be used by the underlying HTTP client.
-   */
-  public GeoApiContext setProxy(Proxy proxy) {
-    requestHandler.setProxy(proxy == null ? Proxy.NO_PROXY : proxy);
-    return this;
+    /**
+     * Disable retries completely.
```
samthor: you might note that this clears the time limit and retries.

My 2c: I find this method a bit hard to grok. Setting retries to zero should be just as effective (even with e.g., a ten hour timeout)—can we just get users to do that?

src/main/java/com/google/maps/model/SpeedLimit.java
```diff
@@ -30,7 +30,7 @@
    */
   public double speedLimit;
 
-  /** Returns the speed limit in miles per hour (MPH). */
+  /** @return Returns the speed limit in miles per hour (MPH). */
```
samthor: Should you rephrase this now?
```@return The speed limit in freedom units (MPH).````

I mean, just drop the 'Returns'.

domesticmouse: I've always written JavaDoc with Returns as the first word of the @return line, because the generated JavaDoc reads a tad weird without it. Or at least it did back in the 1.4 days...

src/main/java/com/google/maps/GeoApiContext.java
domesticmouse: JavaDoc made more explicit as to the actions of the method.

PR #298, opened 2017-07-28
Spell 'adapter' consistently

apjanke:
This project uses a mix of the spellings "adapter" and "adaptor". I assume that is just a typographical quirk, not an intentional distinction. This PR cleans it up by changing everything to "adapter".

I chose "adapter" over "adaptor" because it's the predominant spelling in this project.

```
[~/local/repos/google-maps-services-java on ⇄ master]
$ grep -ri adapter * | wc -l
     182
[~/local/repos/google-maps-services-java on ⇄ master]
$ grep -ri adaptor * | wc -l
       7
```

Google n-grams agrees for general usage:

ngram - adapter vs adaptor - 2017-07-27

This spelling change includes renaming a couple classes, but they're in the `internal` package, so that shouldn't cause any breakage. (Unless clients are intentionally using `internal` classes, which I assume is unsupported.) 

After this change, the build succeeds and the tests pass for me locally, building with JDK 1.8.0_144-b01 on macOS 10.12.6.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 .project
 .classpath
 .vscode/
+/.nb-gradle/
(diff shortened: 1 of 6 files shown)


domesticmouse:
Nice use of data to justify the change. =)

PR #304, opened 2017-08-01
Update GeoApiContext.java - Exception using GAE

gabrielsalmeida:
Fixed GeoApiContext class, because of exception NoClassDefFoundError when using GAE(due of Builder initializing OkHttpRequestHandler). Created another constructor that receives RequestHandler.Builder as workaround.


gabrielsalmeida:
I signed it!

PR #308, opened 2017-08-03
Update build.gradle - Changed guava to v20

gabrielsalmeida:
com.google.guava:guava v22 require JDK 8, changing to guava v20 makes it compatible with JDK 7+.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,7 +52,7 @@ dependencies {
     compile 'joda-time:joda-time:2.9.9'
     compileOnly 'com.google.appengine:appengine-api-1.0-sdk:1.9.54'
     compile 'org.slf4j:slf4j-api:1.7.25'
-    compile 'com.google.guava:guava:22.0'
+    compile 'com.google.guava:guava:20.0'
 
     testCompile 'junit:junit:4.12'
     testCompile 'org.mockito:mockito-core:1.10.19'


domesticmouse:
Sorry to be a pain, but as I have written [elsewhere](https://maps-apis.googleblog.com/2016/09/making-most-of-google-maps-web-service.html) the best way to access the [Google Maps Web Services](https://developers.google.com/maps/web-services/overview) is either via the [Places API for Android](https://developers.google.com/places/android-api/) or via a Server based proxy, which can then utilise one of our [Client Libraries](https://developers.google.com/maps/web-services/client-library), of which this is proudly one.

The reason for requiring this indirection is that embedding the API key for using the Google Maps Web Services into a mobile app, be it Android or iOS opens you up to API key theft. We'd much prefer that not to happen. 

Thanks!

domesticmouse:
Actually, now that I've written that, there are valid use cases for using this library with Java7 in production, so I'll land this PR. But still, please don't use this library in Android. =)

gabrielsalmeida:
Don't worry, I'm not using that for the mobile. I raised that, because I was having problems to make Objectify work with Java 8 runtime (I will check what can be done there to make it compatible).

domesticmouse:
Ahh cool. I'll roll a v0.2.2 release next week to publish this change. Thanks for understanding!

apjanke:
Just another data point: I am using google-maps-services-java with Java 7 in production, and will be for at least another six months. It's being used in a server-side application written in Matlab and using Matlab's embedded JVM, which is currently still on Java 7. I think this counts as one of those valid use cases.

domesticmouse:
I'll roll this fix out soon =)

PR #314, opened 2017-08-09
Javadoc: Update links and fix typos

apjanke:
Updates various outdated/broken links, and fixes some typos and grammatical errors.

Changes are restricted to the Javadocs.

### DIFF
--- a/src/main/java/com/google/maps/DirectionsApiRequest.java
+++ b/src/main/java/com/google/maps/DirectionsApiRequest.java
@@ -83,7 +83,7 @@ public class DirectionsApiRequest
   }
 
   /**
-   * The origin, as a latitude,longitude location.
+   * The origin, as a latitude/longitude location.
    *
    * @param origin The starting location for the Directions request.
    * @return Returns this {@code DirectionsApiRequest} for call chaining.
@@ -93,7 +93,7 @@ public class DirectionsApiRequest
   }
 
   /**
-   * The destination, as a latitude,longitude location.
+   * The destination, as a latitude/longitude location.
    *
    * @param destination The ending location for the Directions request.
    * @return Returns this {@code DirectionsApiRequest} for call chaining.
--- a/src/main/java/com/google/maps/GeocodingApiRequest.java
+++ b/src/main/java/com/google/maps/GeocodingApiRequest.java
@@ -86,11 +86,11 @@ public class GeocodingApiRequest
 
   /**
    * Set the bounding box of the viewport within which to bias geocode results more prominently.
-   * This parameter will only influence, not fully restrict, results from the geocoder. (
+   * This parameter will only influence, not fully restrict, results from the geocoder.
    *
    * <p>For more information see <a
-   * href="https://developers.google.com/maps/documentation/geocoding/?hl=pl#Viewports">Viewports
-   * documentation</a>.
+   * href="https://developers.google.com/maps/documentation/geocoding/intro?hl=pl#Viewports">
+   * Viewport Biasing</a>.
    *
    * @param southWestBound The South West bound of the bounding box.
    * @param northEastBound The North East bound of the bounding box.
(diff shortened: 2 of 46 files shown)


src/main/java/com/google/maps/DirectionsApiRequest.java
```diff
@@ -83,7 +83,7 @@ public DirectionsApiRequest destination(String destination) {
   }
 
   /**
-   * The origin, as a latitude,longitude location.
+   * The origin, as a latitude/longitude location.
```
domesticmouse: /cc @sarahmaddox I'm unsure about this change, thoughts?

sarahmaddox: LGTM

src/main/java/com/google/maps/GeocodingApiRequest.java
```diff
@@ -86,11 +86,11 @@ public GeocodingApiRequest latlng(LatLng latlng) {
 
   /**
    * Set the bounding box of the viewport within which to bias geocode results more prominently.
-   * This parameter will only influence, not fully restrict, results from the geocoder. (
+   * This parameter will only influence, not fully restrict, results from the geocoder.
    *
    * <p>For more information see <a
-   * href="https://developers.google.com/maps/documentation/geocoding/?hl=pl#Viewports">Viewports
-   * documentation</a>.
+   * href="https://developers.google.com/maps/documentation/geocoding/intro?hl=pl#Viewports">
```
sarahmaddox: While you're here, please would you fix this URL? It should be:
https://developers.google.com/maps/documentation/geocoding/start#Viewports

FYI, the URL parameter "hl=pl" shouldn't be there. It's requesting a version of the page in Polish (which we don't actually have) and is probably the result of someone from Poland making this change while viewing the docs.

To see the language specification in action, try this URL which requests the page in German:
https://developers.google.com/maps/documentation/geocoding/start?hl=de

If you leave the parameter off the URL, the developer docs site will look for language info in the user's browser or as specified in the language option at the bottom of the page.

src/main/java/com/google/maps/GeocodingApiRequest.java
```diff
@@ -105,8 +105,8 @@ public GeocodingApiRequest bounds(LatLng southWestBound, LatLng northEastBound)
    * parameter will only influence, not fully restrict, results from the geocoder.
    *
    * <p>For more information see <a
-   * href="https://developers.google.com/maps/documentation/geocoding/?hl=pl#RegionCodes">Region
-   * Codes</a>.
+   * href="https://developers.google.com/maps/documentation/geocoding/intro?hl=pl#RegionCodes">Region
```
sarahmaddox: Similar to my previous comment, the URL should be:
https://developers.google.com/maps/documentation/geocoding/intro#RegionCodes

src/main/java/com/google/maps/GeocodingApiRequest.java
```diff
@@ -120,7 +120,7 @@ public GeocodingApiRequest region(String region) {
    * fully restrict the results from the geocoder.
    *
    * <p>For more information see <a
-   * href="https://developers.google.com/maps/documentation/geocoding/?hl=pl#ComponentFiltering">
+   * href="https://developers.google.com/maps/documentation/geocoding/intro?hl=pl#ComponentFiltering">
```
sarahmaddox: Similar to my previous comment, the URL should be:
https://developers.google.com/maps/documentation/geocoding/intro#ComponentFiltering

src/main/java/com/google/maps/DirectionsApiRequest.java
apjanke: I browsed through the whole source tree, and "latitude/longitude" seems to be a lot more common that "latitude,longitude". Seems like it should have parens and a space if it's going to be "(latitude, longitude)".

src/main/java/com/google/maps/GeocodingApiRequest.java
apjanke: Fixed.

src/main/java/com/google/maps/GeocodingApiRequest.java
apjanke: Fixed.

src/main/java/com/google/maps/GeocodingApiRequest.java
apjanke: Fixed.

apjanke:
Fixed the `hl=pl` things. Sorry for not noticing that earlier.

src/main/java/com/google/maps/DirectionsApiRequest.java
domesticmouse: sgtm

PR #325, opened 2017-08-10
Javadoc: Use "latitude/longitude" instead of "latitude,longitude" everywhere

apjanke:
In this project, when talking about latitude/longitude coordinates, the styling "latitude/longitude" is predominant over "latitude,longitude". (And IMHO, "latitude/longitude" is more conventional English prose.)

```
$ grep -r latitude/longitude * | wc -l
      22
$ grep -r latitude,longitude * | wc -l
       2
```

This PR changes the remaining "latitude,longitude" instances to be "latitude/longitude".

It also removes the `{@code ...}` markup from one of the instances, because it's not referring to a literal code value or identifiers there.

### DIFF
--- a/src/main/java/com/google/maps/DirectionsApiRequest.java
+++ b/src/main/java/com/google/maps/DirectionsApiRequest.java
@@ -190,7 +190,7 @@ public class DirectionsApiRequest
   }
 
   /**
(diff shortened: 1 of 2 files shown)


domesticmouse:
Thanks!

Issue #244, opened 2017-03-21
AutocompletePrediction does not have structured_formatting


dimashi:
This is part of json response that is not copied to AutocompletePrediction:
"structured_formatting": {
        "main_text": "1 Main Street",
        "main_text_matched_substrings": [
          {
            "length": 7,
            "offset": 0
          }
        ],
        "secondary_text": "City, MA, United States"
      }
I was not able to find documentation on what this object represents, but it "looks like" first part is street address, and second is city, state, country, and I need exactly that.

domesticmouse:
Can you please supply a reproduction case that generates this `structured_formatting`?

dimashi:
I believe any autocomplete request returns it. For example this one:
https://maps.googleapis.com/maps/api/place/autocomplete/json?key=???????&input=1 Main Street
```"predictions": [
    {
      "description": "1 Main Street, Acton, MA, United States",
      "id": "3592e3b0cb5087b46f15781773166a8725238b26",
      "matched_substrings": [
        {
          "length": 1,
          "offset": 0
        },
        {
          "length": 11,
          "offset": 2
        }
      ],
      "place_id": "ChIJNZASYb2R44kRVaiUqQsdqoQ",
      "reference": "CkQ_AAAAh7dqgo2Yh9nk_A_B-_PWkkk0v_xeMkVbDowUFKs3s3nAOuwkFLrgqaWD6wDcVI9K2Du4IG6aPk6yqCExBNJZXRIQ1fVBJMRlqR1x83GhWy72JhoUZrXhCWpPDp2hwgQocglNNuxQcUc",
      "structured_formatting": {
        "main_text": "1 Main Street",
        "main_text_matched_substrings": [
          {
            "length": 1,
            "offset": 0
          },
          {
            "length": 11,
            "offset": 2
          }
        ],
        "secondary_text": "Acton, MA, United States"
      },```

domesticmouse:
This will be fixed in v0.2.2

PR #330, opened 2017-08-14
Adding handling for AutocompletePrediction's structured formatting.

domesticmouse:
This is a fix for https://github.com/googlemaps/google-maps-services-java/issues/244

### DIFF
--- a/src/main/java/com/google/maps/model/AutcompleteStructuredFormatting.java
+++ b/src/main/java/com/google/maps/model/AutcompleteStructuredFormatting.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2017 Google Inc. All rights reserved.
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this
+ * file except in compliance with the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+ * ANY KIND, either express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.google.maps.model;
+
+/** Represents the Structured Formatting field of an {@see AutocompletePrediction}. */
+public class AutcompleteStructuredFormatting {
+
+  /** Contains the main text of a prediction, usually the name of the place. */
+  public String mainText;
+
+  /** Contains an array of where the query matched the returned main text. */
+  public AutocompletePrediction.MatchedSubstring mainTextMatchedSubstrings[];
+
+  /** Contains the secondary text of a prediction, usually the location of the place. */
+  public String secondaryText;
+}
--- a/src/main/java/com/google/maps/model/AutocompletePrediction.java
+++ b/src/main/java/com/google/maps/model/AutocompletePrediction.java
@@ -60,16 +60,13 @@ public class AutocompletePrediction {
   }
 
   /**
-   * Describe the location of the entered term in the prediction result text, so that the term can
-   * be highlighted if desired.
+   * The locations of the entered term in the prediction result text, so that the term can be
+   * highlighted if desired.
    */
   public MatchedSubstring matchedSubstrings[];
 
-  /**
-   * Structured formatting contains a description of how the autocomplete query matched the returned
-   * result.
-   */
-  public AutcompleteStructuredFormatting structuredFormatting;
+  /** A description of how the autocomplete query matched the returned result. */
+  public AutocompleteStructuredFormatting structuredFormatting;
 
   /**
    * Identifies each section of the returned description. (A section of the description is generally
--- a/src/main/java/com/google/maps/model/AutocompletePrediction.java
+++ b/src/main/java/com/google/maps/model/AutocompletePrediction.java
@@ -59,7 +59,17 @@ public class AutocompletePrediction {
     public int offset;
   }
 
+  /**
+   * Describe the location of the entered term in the prediction result text, so that the term can
+   * be highlighted if desired.
+   */
   public MatchedSubstring matchedSubstrings[];
+
+  /**
+   * Structured formatting contains a description of how the autocomplete query matched the returned
+   * result.
+   */
+  public AutcompleteStructuredFormatting structuredFormatting;
 
   /**
    * Identifies each section of the returned description. (A section of the description is generally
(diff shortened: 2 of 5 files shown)


src/main/java/com/google/maps/model/AutocompletePrediction.java
```diff
@@ -59,9 +59,19 @@
...
   }
 
+  /**
+   * Describe the location of the entered term in the prediction result text, so that the term can
+   * be highlighted if desired.
+   */
   public MatchedSubstring matchedSubstrings[];
 
   /**
+   * Structured formatting contains a description of how the autocomplete query matched the returned
```
apjanke: I'd simplify to `A description of how the autocomplete query matched the returned result.`. Leading with "Structured formatting contains..." is basically repeating the class name.

src/main/java/com/google/maps/model/AutcompleteStructuredFormatting.java
```diff
@@ -0,0 +1,29 @@
...
+/** Represents the Structured Formatting field of an {@see AutocompletePrediction}. */
+public class AutcompleteStructuredFormatting {
+
+  /** Contains the main text of a prediction, usually the name of the place. */
+  public String mainText;
+
+  /** Contains an array of where the query matched the returned main text. */
+  public AutocompletePrediction.MatchedSubstring mainTextMatchedSubstrings[];
+
+  /** Contains the secondary text of a prediction, usually the location of the place. */
```
apjanke: I'd omit "Contains".

apjanke:
Thanks for the interest!

I made some minor style comments.

There's maybe a substantive issue to cover here, too: Are the `length` and `offset` values in `MatchedSubstring` in Unicode characters, or Java `char`s? Those are often, but not always, the same thing – Java `char`s are UTF-16 code units, not necessarily full Unicode characters/code points, as characters outside the BMP may be represented by surrogate pairs. Normally in all-Java APIs this isn't an issue, but since we're dealing with an external API here, there may be an "impedance mismatch". I've seen other places in the Google Maps API docs where it describes offsets as being "in Unicode characters". If that's the case here too, it's worth noting. If `MatchedSubstring` is carrying Unicode-character offsets, instead of `char` offsets, that might be a surprise to your average Java developer. May be worth noting in the Javadoc which it is.

(I don't know how likely this difference is to crop up in practice for Google Maps, because I don't know how far its localization reaches, but it might be worth noting for full correctness.)

domesticmouse:
The APIs are served in UTF8 I believe, and the length and offsets should be Unicode characters, but I'm unsure whether we document this anywhere. 

At this point I'm copying across the documentation from the Places API documentation, as that is what is published, and I'm uncomfortable going above and beyond that.

That said, if someone tests this, and finds that we aren't serving length and offset as unicode characters, then please raise a bug on our [issue tracker](https://developers.google.com/maps/support/#issue_tracker). 

This client library is merely passing through what the APIs return, if the APIs are returning data that is incorrect then we have to fix the APIs themselves, not this client library.

src/main/java/com/google/maps/model/AutcompleteStructuredFormatting.java
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+ * ANY KIND, either express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.google.maps.model;
+
+/** Represents the Structured Formatting field of an {@see AutocompletePrediction}. */
+public class AutcompleteStructuredFormatting {
```
domesticmouse: Done, and yes, that's quite a typo right there...

src/main/java/com/google/maps/model/AutcompleteStructuredFormatting.java
```diff
@@ -0,0 +1,29 @@
...
+
+  /** Contains the main text of a prediction, usually the name of the place. */
```
domesticmouse: Done

src/main/java/com/google/maps/model/AutcompleteStructuredFormatting.java
```diff
@@ -0,0 +1,29 @@
...
+  public String mainText;
+
+  /** Contains an array of where the query matched the returned main text. */
```
domesticmouse: done

src/main/java/com/google/maps/model/AutcompleteStructuredFormatting.java
domesticmouse: done

src/main/java/com/google/maps/model/AutocompletePrediction.java
```diff
@@ -59,9 +59,19 @@
     public int offset;
   }
 
+  /**
+   * Describe the location of the entered term in the prediction result text, so that the term can
```
domesticmouse: done

src/main/java/com/google/maps/model/AutocompletePrediction.java
domesticmouse: done

apjanke:
Oh, I don't think it's an issue of correctness with the APIs, just a matter of setting expectations for Java-centric developers who use this library. From what I've seen, most Java APIs seem to use Java `char` counts for string offsets, and mostly overlook the fact that Java `char`s are not exactly equivalent to Unicode characters. That is, coders expect offsets and lengths to index in to `String`s using the `charAt()` method or array indexing into the corresponding `char[]` array, and may need to be given notice that they should be using the `offsetByCodePoints()` method instead. I think simply noting in the Javadoc for those fields that they are offsets in Unicode characters (as opposed to Java `char`s) would be sufficient and appropriate. Like the existing documentation for `AutocompletePrediction.terms` does [here](https://github.com/googlemaps/google-maps-services-java/blob/1616f4f2c16c24dc4db844945f4905133760514f/src/main/java/com/google/maps/model/AutocompletePrediction.java#L45); I'd just add "`...measured in Unicode characters` to the new length/offset fields' Javadoc.

Sorry if I'm belaboring a point that's already been gone over here in the past.

domesticmouse:
Valid commentary, I'm not totally up to speed on I18N concerns, so I value this feedback. And I'd totally forgotten that Java is UTF16. I've obviously spent too long with Go and it's Unicode runes....

domesticmouse:
How's that look @apjanke?

apjanke:
Yep, LGTM!

domesticmouse:
Thanks for your careful reviews @apjanke!

Issue #338, opened 2017-08-18
Rate limiting not working correctly


dvankley:
I've been trying out the library for making nearby search requests lately and overall it works great. I am having a timing issue that I haven't been able to clear up, namely that I keep getting `InvalidRequestExceptions` on my `nearbySearchNextPage()` calls following my initial `nearbySearchQuery()` call. If I add a `Thread.sleep()` before the `nearbySearchQuery` call then I don't see the error, which is what makes me think it's a rate limiting issue of some kind. I'm on version 0.2.1

It doesn't seem to be on Google's end because I don't see any errors in the API console.

Manually setting the rate limit very low or high with `queryRateLimit()` also doesn't help.

Any hints on what I'm doing wrong would be helpful. Thanks!

Sample code:
```
    public static void main( String[] args )
    {
        try {
            String apiKey = "Insert API Key here";

            Double homeLat = 33.7606946;
            Double homeLong = -84.3907441;

            GeoApiContext.Builder builder = new GeoApiContext.Builder();
            GeoApiContext context = builder
                    .apiKey(apiKey)
                    .build();

            NearbySearchRequest request = PlacesApi.nearbySearchQuery(context, new LatLng(homeLat, homeLong));
            request.radius(5000);
            request.type(PlaceType.RESTAURANT);

            PlacesSearchResponse response = request.await();

            while (response.nextPageToken != null) {
                request = PlacesApi.nearbySearchNextPage(context, response.nextPageToken);
                request.radius(1000);
                request.type(PlaceType.RESTAURANT);
                response = request.await();
            }
        } catch (Exception e) {
            System.out.println("Exception: " + e.getMessage());
        }
    }
```

domesticmouse:
Per https://developers.google.com/places/web-service/search#PlaceSearchPaging there is a short delay between the result containing the `next_page_token` and the page token becoming valid.

dvankley:
Makes sense, thanks. Mind if I add a note to the Readme here to remind users of that?

PR #339, opened 2017-08-18
Add note about paging delay

dvankley:
From discussion in https://github.com/googlemaps/google-maps-services-java/issues/338


dvankley:
I signed your thing

PR #365, opened 2017-11-05
added a shutdown method to GeoApiContext which stops RateLimitExecutorDelayThread #261

stafichuk:
This is a pull request for issue #261. Details can be found in the [comment](https://github.com/googlemaps/google-maps-services-java/issues/261#issuecomment-341965630) but I will also copy them here for convenience.

The root cause of the issue is following. We are creating and starting a thread in a constructor of class [RateLimitExecutorService](https://github.com/googlemaps/google-maps-services-java/blob/7d7f0b998a734acc7f32ffb4760e84c1d288e4bb/src/main/java/com/google/maps/internal/RateLimitExecutorService.java#L36). Here it is:
```java
public RateLimitExecutorService() {
    setQueriesPerSecond(DEFAULT_QUERIES_PER_SECOND);
    Thread delayThread = new Thread(this);
    delayThread.setDaemon(true);
    delayThread.setName("RateLimitExecutorDelayThread");
    delayThread.start();
}
```
[run](https://github.com/googlemaps/google-maps-services-java/blob/7d7f0b998a734acc7f32ffb4760e84c1d288e4bb/src/main/java/com/google/maps/internal/RateLimitExecutorService.java#L72) method of this class looks like this:
```java
@Override
public void run() {
  try {
    while (!delegate.isShutdown()) {
      this.rateLimiter.acquire();
      Runnable r = queue.take();
      delegate.execute(r);
    }
  } catch (InterruptedException ie) {
    LOG.info("Interrupted", ie);
  }
}
```
There are two problems with this method:
1. `RateLimitExecutorDelayThread` should end when the delegate is shut down but the thing is [RateLimitExecutorService#shutdown](https://github.com/googlemaps/google-maps-services-java/blob/7d7f0b998a734acc7f32ffb4760e84c1d288e4bb/src/main/java/com/google/maps/internal/RateLimitExecutorService.java#L103) method and thus `shutdown()` method of the delegate is never called.
2. If we call `shutdown()` method but the `queue` is empty we will wait forever in `queue.take()`.

This issue manifests itself when you are restarting an application multiple times inside a single JVM without restarting JVM itself. For example, it can happen in application servers like Tomcat or in Play Framework's dev mode, which was my case.

### DIFF
--- a/src/main/java/com/google/maps/GaeRequestHandler.java
+++ b/src/main/java/com/google/maps/GaeRequestHandler.java
@@ -93,6 +93,11 @@ public class GaeRequestHandler implements GeoApiContext.RequestHandler {
         req, client, clazz, fieldNamingPolicy, errorTimeout, maxRetries, exceptionsAllowedToRetry);
   }
 
+  @Override
+  public void shutdown() {
+    //do nothing
+  }
+
   /** Builder strategy for constructing {@code GaeRequestHandler}. */
   public static class Builder implements GeoApiContext.RequestHandler.Builder {
 
--- a/src/main/java/com/google/maps/GeoApiContext.java
+++ b/src/main/java/com/google/maps/GeoApiContext.java
@@ -104,6 +104,8 @@ public class GeoApiContext {
         Integer maxRetries,
         ExceptionsAllowedToRetry exceptionsAllowedToRetry);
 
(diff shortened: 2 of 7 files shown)


stafichuk:
I signed it!

Issue #366, opened 2017-11-06
Cache - Class model not implement java.io.Serializable


lfochi:
If you want implements a Cache with the same response model is not posible because it does not implement java.io.Serializable

domesticmouse:
Please feel free to create a pull request that adds Serializable to the response model. Thanks!

domesticmouse:
This will show up in [v0.2.5](https://github.com/googlemaps/google-maps-services-java/releases/tag/v0.2.5)

lfochi:
Thanks @domesticmouse !

PR #367, opened 2017-11-07
added java.io.Serializable to the response model #366

lfochi:
This is a pull request for issue #366

### DIFF
--- a/src/main/java/com/google/maps/model/AddressComponent.java
+++ b/src/main/java/com/google/maps/model/AddressComponent.java
@@ -15,6 +15,7 @@
 
 package com.google.maps.model;
 
+import java.io.Serializable;
 /**
  * The parts of an address.
  *
@@ -23,7 +24,10 @@ package com.google.maps.model;
  * href="https://developers.google.com/maps/documentation/geocoding/intro">Google Maps Geocoding API
  * Developer's Guide</a> for more detail.
  */
-public class AddressComponent {
+public class AddressComponent implements Serializable {
+  /** The default serialVersionUID */
+  private static final long serialVersionUID = 1L;
+
   /** The full text description or name of the address component as returned by the Geocoder. */
   public String longName;
 
--- a/src/main/java/com/google/maps/model/AddressComponent.java
+++ b/src/main/java/com/google/maps/model/AddressComponent.java
@@ -25,7 +25,7 @@ import java.io.Serializable;
  * Developer's Guide</a> for more detail.
  */
 public class AddressComponent implements Serializable {
-  /** The default serialVersionUID */
+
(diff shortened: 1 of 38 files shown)


lfochi:
I signed it!

src/main/java/com/google/maps/model/AddressComponent.java
```diff
@@ -23,7 +24,10 @@
  * href="https://developers.google.com/maps/documentation/geocoding/intro">Google Maps Geocoding API
  * Developer's Guide</a> for more detail.
  */
-public class AddressComponent {
+public class AddressComponent implements Serializable {
+  /** The default serialVersionUID */
+  private static final long serialVersionUID = 1L;
```
domesticmouse: Hey @markmcd and @samthor, is this a legit way of introducing `serialVersionUID` for a class that is just now implementing `Serializable`? I believe this shouldn't break anyone, but I'd like confirmation... =)

samthor: I think this is fine. It only mattered if `AddressComponent` was previously `Serializable`, which was not.

FWIW, my limited reading suggests that the reason most `serialVersionUID` fields are random super high numbers is that IDEs will create them using the previous default Java hashing implementation. But again, as the class was never serializable, anything is fine here.

src/main/java/com/google/maps/model/AddressComponent.java
```diff
@@ -23,7 +24,10 @@
  * href="https://developers.google.com/maps/documentation/geocoding/intro">Google Maps Geocoding API
  * Developer's Guide</a> for more detail.
  */
-public class AddressComponent {
+public class AddressComponent implements Serializable {
+  /** The default serialVersionUID */
```
samthor: I think this comment is probably redundant (on all classes).

src/main/java/com/google/maps/model/AddressComponent.java
domesticmouse: Thanks Sam, I concur.

src/main/java/com/google/maps/model/AddressComponent.java
lfochi: Thanks Sam, I concur but as this was the first time we introduce Serializable, I thought that this comment was neccessary. But I also agree with you that probably it's redundant in every class.
Should I delete them before you aprobed this pull request? 
I tried to make real notice that that field was introduced for the serializable interface

domesticmouse: @lfochi please remove, and then I'll merge. Thanks!

domesticmouse:
Thanks @lfochi !

PR #389, opened 2017-11-23
Add java.io.Serializable for inner classes, as it was only added in t…

chufaca:
…op-level ones.

This is needed to implement a cache of the model classes.

Thanks.


chufaca:
I signed it!

domesticmouse:
Thanks!

PR #393, opened 2017-12-04
Made the constructor public

Sameer9:
Rather than making the DirectionApiRequest.java as singleton class, we need to give its constructor the public access.

PR #394, opened 2017-12-05
Revert "Made the constructor public"

domesticmouse:
Reverts googlemaps/google-maps-services-java#393

Change not under CLA


domesticmouse:
@googlebot this is reverting out a change that wasn't correctly CLA'd.

PR #409, opened 2018-03-22
close response on retry

esetnik:
This fixes the issue reported in https://github.com/googlemaps/google-maps-services-java/issues/179


esetnik:
I've been running the alternative version in our production codebase for the last two weeks without a recurrence of #179

According to the [ResponseBody documentation](https://square.github.io/okhttp/3.x/okhttp/okhttp3/ResponseBody.html)

```
For synchronous calls, the easiest way to make sure a response body is closed is with a try block. With this structure the compiler inserts an implicit finally clause that calls close() for you.

   

   Call call = client.newCall(request);
   try (Response response = call.execute()) {
     ... // Use the response.
   }

```
Which is closed on https://github.com/googlemaps/google-maps-services-java/blob/master/src/main/java/com/google/maps/internal/OkHttpPendingResult.java#L225. But if the code short circuits due to `shouldRetry(response)` returning true, then the response is never closed.

domesticmouse:
Good catch, thanks!

Issue #233, opened 2017-03-05
Google Static Maps API support


IvanRF:
I think this library can easily support the [Google Static Maps API](https://developers.google.com/maps/documentation/static-maps/intro) too. Most of the communication protocols are already implemented, like the `UrlSigner` class for [Get a Key and Signature/Authentication](https://developers.google.com/maps/documentation/static-maps/get-api-key).

The only work to do is to map the parameters that can be included in a call to `https://maps.googleapis.com/maps/api/staticmap?...`.

I'm already using both this library and HTTP calls to the Static Maps API to get the `byte[]` of a map's image, which I then convert to a `BufferedImage`.

According to the documentation, Google Static Maps API is listed for the [Web platform](https://developers.google.com/maps/web/) and this library is intended for [Web services](https://developers.google.com/maps/web-services/). Nevertheless, this specific API can be useful for Java applications too, not only webs.

domesticmouse:
This will be fixed in v0.2.2.

domesticmouse:
Actually, I was thinking of the golang implementation. Whups.

IvanRF:
@domesticmouse first of all, thank you very much for adding Static Maps API to the library! 👍 

I just replaced my own implementation and started using the library for Static Maps API calls too. So far everything seems to work fine.

The only thing that I was using and it was not implemented here is `Encoded polyline` as a valid value for `path`.

From https://developers.google.com/maps/documentation/maps-static/dev-guide

> path (optional) defines a single path of two or more connected points to overlay on the image at specified locations. This parameter takes a string of point definitions separated by the pipe character (|), **or an encoded polyline using the `enc:` prefix within the location declaration of the path**.

More info at: [EncodedPolylines](https://developers.google.com/maps/documentation/maps-static/dev-guide#EncodedPolylines).

I was using this in the URL in order to show a map from a `DirectionsRoute`:
```
DirectionsRoute route;
...
EncodedPolyline polyline = route.overviewPolyline;
parameters += "&path=enc" + colon + URLEncoder.encode(polyline.getEncodedPath(), "UTF-8");
```

Now with the library I'm using this:
```
Path path = new Path();
for (LatLng point : polyline.decodePath())
	path.addPoint(point);
req.path(path);
```

apjanke:
Yeah, that seems a little verbose and inconvenient.

[...]

Issue #412, opened 2018-04-03
Add library to enum AddressType


Makabane:
Hi, I got this warning : 
`Unknown type for enum com.google.maps.model.AddressType: 'library'`
when I request 
> https://maps.googleapis.com/maps/api/geocode/json?key=[...]&address=80+FR
with google-maps-services-0.1.15.jar

domesticmouse:
This will be in the next release, which should go out shortly. Thanks!

PR #413, opened 2018-04-08
Adding Static Maps API

domesticmouse:
Fixes https://github.com/googlemaps/google-maps-services-java/issues/233. 

Also contains random clean ups and fixes, e.g. https://github.com/googlemaps/google-maps-services-java/issues/412


src/main/java/com/google/maps/StaticMapsRequest.java
```diff
@@ -0,0 +1,415 @@
...
+    public void size(MarkersSize size) {
+      this.size = size;
+    }
+
+    /**
+     * Specifies a 24-bit color (example: color=0xFFFFCC) or a predefined color from the set {black,
+     * brown, green, purple, yellow, blue, gray, orange, red, white}.
+     */
+    public void color(String color) {
+      if (!colorPattern.matcher(color).matches()) {
```
samthor: is it worth enforcing this locally? We might allow more colors in future.

src/main/java/com/google/maps/StaticMapsRequest.java
```diff
@@ -0,0 +1,415 @@
...
+        return this.name();
+      }
+    }
+
+    private MarkersSize size;
+    private String color;
+    private String label;
+    private String customIconURL;
+    private CustomIconAnchor anchorPoint;
+    private List<String> locations = new ArrayList<>();
```
samthor: private final, you don't ever change this ref

src/main/java/com/google/maps/StaticMapsRequest.java
```diff
@@ -0,0 +1,415 @@
...
+    return param("markers", markers);
+  }
+
+  public static class Path implements UrlValue {
+
+    private Integer weight;
+    private String color;
+    private String fillcolor;
+    private boolean geodesic;
+    private List<String> points = new ArrayList<>();
```
samthor: private final

src/main/java/com/google/maps/StaticMapsRequest.java
```diff
@@ -0,0 +1,415 @@
...
+
+    /**
+     * Specifies the thickness of the path in pixels. If no weight parameter is set, the path will
+     * appear in its default thickness (5 pixels).
+     */
+    public void weight(int weight) {
+      this.weight = weight;
```
samthor: should setting this to zero clear the Integer?

src/main/java/com/google/maps/model/Size.java
```diff
@@ -0,0 +1,38 @@
...
+   *
+   * @param height The height of this Size.
+   * @param width The width of this Size.
+   */
+  public Size(int width, int height) {
+    this.width = width;
+    this.height = height;
+  }
+
+  /** Serialisation constructor. */
```
samthor: american spelling, s => z

src/test/java/com/google/maps/StaticMapsApiTest.java
```diff
@@ -0,0 +1,161 @@
...
+import com.google.maps.StaticMapsRequest.Markers;
+import com.google.maps.StaticMapsRequest.Markers.CustomIconAnchor;
+import com.google.maps.StaticMapsRequest.Markers.MarkersSize;
+import com.google.maps.StaticMapsRequest.Path;
+import com.google.maps.StaticMapsRequest.StaticMapType;
+import com.google.maps.model.LatLng;
+import com.google.maps.model.Size;
+import java.awt.image.BufferedImage;
+import java.io.ByteArrayInputStream;
+import javax.imageio.ImageIO;
```
samthor: Is this library only pulled in for testing? Do we have it for the regular lib?

src/main/java/com/google/maps/PendingResultBase.java
```diff
@@ -96,8 +96,18 @@ protected A param(String key, String val) {
     return result;
   }
 
+  protected A param(String key, int val) {
+    params.put(key, Integer.toString(val));
+
+    @SuppressWarnings("unchecked") // safe by specification - A is the actual class of this instance
```
samthor: minor nit: could you replace these three returns with

```java
private A getInstance() {
  @SuppressWarnings("unchecked");
  return (A) this;
}

// later
return this.getInstance();
```

domesticmouse: I like it, done. Ish.

src/main/java/com/google/maps/ImageResult.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public ImageResult(String contentType, byte[] imageData) {
+    this.imageData = imageData;
+    this.contentType = contentType;
+  }
+
+  private static final long serialVersionUID = 1L;
+
+  /** The image data from the Photos API call. */
+  public final byte[] imageData;
```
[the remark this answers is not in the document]
domesticmouse: This `ImageResult` is an output value object. If a client of this library alters the image data, it has no impact on the library.

src/main/java/com/google/maps/StaticMapsRequest.java
domesticmouse: Done

src/main/java/com/google/maps/StaticMapsRequest.java
domesticmouse: Fair call, ripped out regex

src/main/java/com/google/maps/StaticMapsRequest.java
domesticmouse: done

src/main/java/com/google/maps/StaticMapsRequest.java
domesticmouse: Good idea. Implemented in `toUrlValue()`

src/main/java/com/google/maps/model/Size.java
domesticmouse: done

src/test/java/com/google/maps/StaticMapsApiTest.java
domesticmouse: I only use it in the test library to make sure we can round trip an image. I'm not decoding the image in the main library to keep up throughput.

src/main/java/com/google/maps/PendingResultBase.java
```diff
@@ -89,32 +91,32 @@ public final void cancel() {
   protected abstract void validateRequest();
 
   protected A param(String key, String val) {
-    params.put(key, val);
+    if(params.get(key) == null) {
```
samthor: nit: missing space `if (`

src/main/java/com/google/maps/PendingResultBase.java
```diff
@@ -89,32 +91,32 @@ public final void cancel() {
...
+      params.put(key, new ArrayList<String>());
```
samthor: This looks naïvely fine but I wonder if anyone is using this class to overwrite previous values being set via `.param()`.

Would it be worth having an explicit 'add' flag somehow? I know you already have the three `param` methods, so a matrix of those would be awkward. Let me know.

src/test/java/com/google/maps/LocalTestServerContext.java
```diff
@@ -117,6 +117,21 @@ void assertParamValue(String expectedValue, String paramName)
...
+      this.params = this.actualParams();
+    }
+    int paramsFound = 0;
+    for (NameValuePair pair : params) {
+      if (pair.getName().equals(paramName)) {
+        assertEquals(expectedValues.get(paramsFound), pair.getValue());
+        paramsFound++;
+      }
+    }
+    assertTrue(paramsFound == expectedValues.size());
```
samthor: why not assertEquals?

domesticmouse: Good catch, fixed.

src/main/java/com/google/maps/PendingResultBase.java
domesticmouse: In fact, yes I was being naive. There was logic in optimise waypoints that implicitly relied on that underlying hashmap behavior of the parameters. 

New implementation now enforces singleton parameters (as implicitly required by non Static Maps API surfaces), and just uses the multiple parameters for Static Maps `path`s and `markers`.

Now the tests pass...

PR #419, opened 2018-04-30
Add support for scale parameter on custom icons to be able to use hi-DPI icons when map scale > 1

astachelek:
It appears that Google supports a `scale` parameter for custom marker icons although this is undocumented. See the Stack Overflow articles below. This is ultimately necessary when generating maps for hi-DPI screens where the overall map `scale` option is set to 2 or 4. I've tested with this locally and it appears to be working albeit it is undocumented by Google.

https://stackoverflow.com/questions/8813143/how-do-i-add-a-high-resolution-custom-marker-on-a-static-google-map

https://stackoverflow.com/questions/10336646/how-can-i-use-high-resolution-custom-markers-with-the-scale-parameter-in-google

### DIFF
--- a/src/main/java/com/google/maps/StaticMapsRequest.java
+++ b/src/main/java/com/google/maps/StaticMapsRequest.java
@@ -16,10 +16,12 @@
 package com.google.maps;
 
 import com.google.maps.internal.ApiConfig;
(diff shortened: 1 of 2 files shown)


astachelek:
I signed it!

PR #421, opened 2018-05-14
Migrate to java.time

ptahchiev:
The PR is ok, however the build tells me files are not formatted properly - spent a few hours but I gave up. Can you please format them properly?


domesticmouse:
The way to fix formatting is to run `./gradlew googleJavaFormat` =)

domesticmouse:
I'm going to hold this PR for six months because we still have developers on Java 1.7. Sorry about that.

ptahchiev:
No one said the project will switch all the versions automatically to 1.8. You will still be able to use the previously released versions which you are probably using right now. And if the developers are kind enough they may even support separate branches for the different java versions.

apjanke:
> You will still be able to use the previously released versions which you are probably using right now. 

Yes, but I'm interested in keeping up with the latest versions. Because `google-maps-services-java` uses an enum to represent addresss types, and new address types are added on a regular basis, it's somewhat important for me to keep up with the latest version. This is different from most Java projects, where data changes do not imply code changes so much.

> And if the developers are kind enough they may even support separate branches for the different java versions.

I wouldn't ask that of them. That seems like a pain. :)

domesticmouse:
@apjanke Any feel for how much longer you are going to be stuck on 1.7? I'd prefer not to split this client library if possible =)

apjanke:
> Any feel for how much longer you are going to be stuck on 1.7?

Until close to the end of the year, I'm afraid.

Don't split the library on my account: I can always get by with one of the older versions unless the API changes; it's just an inconvenience. In fact, if this library goes Java 1.8, that would help me make the argument that we should upgrade to Java 8 sooner. And regardless, I can't really argue that this library should stick on Java 1.7, because 1.7 is already end-of-life.

domesticmouse:
I've got other people in the same position, thus my decision to hold for six months. I'll stay on that timeline, and now you have leverage in the form of a known deadline. Hope this helps.

apjanke:
It does. Thanks!

domesticmouse:
Hey all, I'd like to land this change next month, given we are about six months after when I'd said I'd hold this change for six months. How is everyone situated? @apjanke ?

@ptahchiev any chance of bringing this change up to date so I can land it easily? Thanks!

apjanke:
I say go for it. I'm still stuck on Java 1.7 at work, but that's no reason to hold the community back. 1.7's time is over.

domesticmouse:
Thanks @apjanke =)

PR #437, opened 2018-06-29
Add more AddressTypes and AddressComponentTypes

apjanke:
This PR adds more currently-unrecognized AddressComponentTypes and AddressTypes.

AddressTypes:
* general_contractor

AddressComponentTypes:
* travel_agency
* electronics_store
* home_goods_store
* school
* store
* art_gallery
* lawyer
* restaurant
* bar
* meal_takeaway
* clothing_store
* local_government_office
* finance
* moving_company
* storage

These were discovered using my [gmsj-cli](https://github.com/apjanke/gmsj-cli) fuzz tester.

```
$ ./bin/gmsj-cli geofuzz "555 Broadway, New York, NY" -n 5000 -S 123456                                                                                           master
Searching 5000 points around '555 Broadway, New York, NY', with radius 0.100000 degrees (rand seed 123456)
[main] WARN com.google.maps.internal.SafeEnumAdapter - Unknown type for enum com.google.maps.model.AddressComponentType: 'real_estate_agency'
[main] WARN com.google.maps.internal.SafeEnumAdapter - Unknown type for enum com.google.maps.model.AddressComponentType: 'travel_agency'
[main] WARN com.google.maps.internal.SafeEnumAdapter - Unknown type for enum com.google.maps.model.AddressComponentType: 'real_estate_agency'
[main] WARN com.google.maps.internal.SafeEnumAdapter - Unknown type for enum com.google.maps.model.AddressComponentType: 'travel_agency'
[...]
```

PR #441, opened 2018-06-29
Add AddressComponentType insurance_agency

apjanke:
This PR adds the "insurance_agency" AddressComponentType. It's present as a PlaceType and AddressType, but not an AddressComponentType, and in testing, I've seen it show up in AddressComponentType.


apjanke:
Merged into #437.

PR #452, opened 2018-07-02
Add value-based toString()s for model objects

### LINKED ISSUE #450: Value-based toString() for model objects?
Would you be interested in a PR that added value-based `toString()` implementations to the classes in `com.google.maps.model`?

Right now, most of them use the default `java.lang.Object` `toString()`, which returns something like `<ClassName>@<hashCode>`, and since they use the default `hashCode()`, that gives a value based on object identity.

These are value classes, so their content is probably more important than object identity. For example, `Bounds` could be represented as `[(<ne-lat>,<ne-lng>),(<sw-lat>,<sw-lng>)]` instead of `Bounds@8713648763`.

This could be useful for incorporating objects into log messages, or for interactive use in scripting environments like Matlab or Jython. Particularly for log messages: object identity is not significant outside of the current process, so the default `Object.toString()` representation is probably not very useful in logs.

apjanke:
Closes #450.

This PR adds value-based custom `toString()` methods to all the `model` objects. This should be useful for debugging, including incorporation into log messages, and interactive use inside scripting environments like Matlab and Jython.

Value-based toString()s are appropriate here because the model objects are value objects; their object identity doesn't matter as much as their contents, so the default `<ClassName>@<ObjectId>` toString() output is not very useful.

This PR affects user-visible behavior for existing classes, so you might want to sit on it a few days to allow other users to comment.

If this approach looks good to you, I'll add some tests.

### DIFF
--- a/src/main/java/com/google/maps/model/EncodedPolyline.java
+++ b/src/main/java/com/google/maps/model/EncodedPolyline.java
@@ -50,4 +50,10 @@ public class EncodedPolyline implements Serializable {
   public List<LatLng> decodePath() {
     return PolylineEncoding.decode(points);
   }
+
+  // Use the encoded point representation; decoding to get an alternate representation for
+  // individual points would be expensive.
+  public String toString() {
+    return String.format("[EncodedPolyline: %s]", points);
+  }
 }
--- a/src/main/java/com/google/maps/internal/StringJoin.java
+++ b/src/main/java/com/google/maps/internal/StringJoin.java
@@ -15,6 +15,8 @@
 
 package com.google.maps.internal;
 
+import java.util.Objects;
+
 /** Utility class to join strings. */
 public class StringJoin {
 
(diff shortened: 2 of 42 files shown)


src/main/java/com/google/maps/model/EncodedPolyline.java
```diff
@@ -50,4 +50,7 @@ public String getEncodedPath() {
   public List<LatLng> decodePath() {
     return PolylineEncoding.decode(points);
   }
+
+  // Stick with the default toString; decoding the polyline to get a point count or representation
```
domesticmouse: I'd suggest adding a toString that includes the encoded polyline.

apjanke: Done.

domesticmouse:
I'm still waiting on tests for this PR, amirite?

apjanke:
Yes. Sorry, been busy with the day job. :)

domesticmouse:
All good =)

domesticmouse:
Looks like I broke your PR. Once this is in I'll roll a release. Thanks!

domesticmouse:
Thanks Andrew, no rush

apjanke:
Added tests. And what do you know, they caught a bug in one of my toString()s.

The tests work by just running `toString()` on the "live" data used in the existing query tests.

The tests don't check the content of the resulting string. Since these are human-readable strings for debugging, and don't have a fixed format or direct relation to an API, I don't think testing their content is necessary, or even appropriate.

If you want, could do more thorough and isolated tests by just directly constructing various permutations of the model objects, including ones with `null` in all the nullable fields, and do toString() on those. Seems a little overkill to me.

domesticmouse:
Looks good to me. 👍

apjanke:
Excellent.

Want to grab #463 before the release, too?

PR #459, opened 2018-07-03
Add Plus Code support to GeocodingResult and PlaceDetails

### LINKED ISSUE #458: plus_code is missing in model GeocodingResult
[plus_code](https://developers.google.com/maps/documentation/geocoding/intro#pluscode) needs to be included in model class [GeocodingResult.java](https://github.com/googlemaps/google-maps-services-java/blob/master/src/main/java/com/google/maps/model/GeocodingResult.java)

apjanke:
Fixes #458.

I've added test data for the PlaceDetails support, but not for the GeocodingResult, because I'm not able to see `plus_code` in my Geocoding queries yet, probably due to a staged rollout of the Plus Code feature. This PR adds forward-compatible support for clients who can see it, though.

### DIFF
--- a/src/main/java/com/google/maps/model/PlusCode.java
+++ b/src/main/java/com/google/maps/model/PlusCode.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2018 Google Inc. All rights reserved.
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this
+ * file except in compliance with the License. You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/maps/model/PlusCode.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright 2015 Google Inc. All rights reserved.
```
domesticmouse: I feel like this should be 2018 =)

apjanke: Amended. :)

PR #468, opened 2018-07-07
Add non-stopover waypoint support

### LINKED ISSUE #93: Allow Non-Stopover Waypoints
From https://developers.google.com/maps/documentation/directions/ -->

You can use waypoints to influence your route without adding a stopover by prefixing the waypoint with via:. Waypoints prefixed with via: will not add an additional leg to the route.

Sample Request:

```
https://maps.googleapis.com/maps/api/directions/json?origin=Boston,MA&destination=Concord,MA&waypoints=Charlestown,MA|via:Lexington,MA&key=API_KEY
```

apjanke:
This PR adds support for non-stopover waypoints.

Fixes #93.

NOTE: This PR also slightly changes the current behavior of the `waypoints(...)` methods. Previously, when `waypoints(...)` was called with a `null` or empty array argument, it was simply ignored. That means that if `waypoints(...)` had previously been called, the effect of the previous value was simply kept. I think that's the wrong behavior: for empty array arguments, and I think for null arguments too if we allow those, I think calling `waypoints([])` or waypoints(null)` should instead reset the waypoints parameter to the empty set. This PR changes the behavior to do so.

This PR also fixes a typo in the function name of one of the tests in DirectionsApiTest.

### DIFF
--- a/src/main/java/com/google/maps/DirectionsApiRequest.java
+++ b/src/main/java/com/google/maps/DirectionsApiRequest.java
@@ -16,6 +16,7 @@
 package com.google.maps;
 
 import static com.google.maps.internal.StringJoin.join;
+import static java.util.Objects.requireNonNull;
 
 import com.google.maps.model.DirectionsResult;
 import com.google.maps.model.LatLng;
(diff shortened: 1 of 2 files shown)


domesticmouse:
Thank you Andrew!

apjanke:
You're welcome!

PR #471, opened 2018-07-07
Replace "baseUrlForTesting()" with usage-neutral "baseUrlOverride()"

### LINKED ISSUE #435: Inappropriate method name : baseUrlForTesting can be used for Google Maps China
[GeoApiContext.java](https://github.com/googlemaps/google-maps-services-java/blob/master/src/main/java/com/google/maps/GeoApiContext.java) names a method as baseUrlForTesting
`Builder baseUrlForTesting(String baseUrl)`
As per [Google Maps China documentation](https://developers.google.com/maps/documentation/javascript/localization#GoogleMapsChina),
this method can also be used for, overriding base Url.
Please consider method name change, or provide overridden method, or mark this method as deprecated and need to improve documentation comment.

apjanke:
Fixes #435.

The name `baseUrlForTesting` implies a certain use case, but the action is really a general override of the base URL. Turns out it's useful in other scenarios, such as querying Google Maps from China, as stated in #435.

This PR replaces `baseUrlForTesting(...)` with a new usage-neutral-terminology `baseUrlOverride(...)` that describes what it does, but not why. The old `baseUrlForTesting(...)` is left around in a `@Deprecated` state for now.

### DIFF
--- a/src/main/java/com/google/maps/GeoApiContext.java
+++ b/src/main/java/com/google/maps/GeoApiContext.java
@@ -339,14 +339,28 @@ public class GeoApiContext {
     }
 
     /**
-     * Overrides the base URL of the API endpoint. Useful only for testing.
+     * Overrides the base URL of the API endpoint. Useful for testing or certain international
+     * usage scenarios.
      *
      * @param baseUrl The URL to use, without a trailing slash, e.g. https://maps.googleapis.com
      * @return Returns this builder for call chaining.
      */
-    Builder baseUrlForTesting(String baseUrl) {
+    Builder baseUrlOverride(String baseUrl) {
       baseUrlOverride = baseUrl;
       return this;
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/maps/GeoApiContext.java
```diff
@@ -339,16 +339,30 @@ public Builder requestHandlerBuilder(RequestHandler.Builder builder) {
     }
 
     /**
-     * Overrides the base URL of the API endpoint. Useful only for testing.
+     * Overrides the base URL of the API endpoint. Useful for testing or certain international
+     * usage scenarios, such as querying the API from China.
```
domesticmouse: Can you please delete the reference to China in this doc comment? Thanks!

apjanke: Done!

domesticmouse:
Thanks!

PR #481, opened 2018-07-17
Remove Radar-related test resource files

apjanke:
Radar search support has been deprecated and disabled, and the Radar
support has already been removed from our code in #469. This just
cleans up the remaining test support files.


domesticmouse:
Thanks for the clean up 😇

PR #506, opened 2018-10-19
Fixed typo in field

githubsaturn:
According to the official docs, the field should be called "review", not "reviews". It appears that "reviews" also work, but the correct term is "review". More info in the doc:
https://developers.google.com/places/web-service/details#fields


githubsaturn:
I signed it!

domesticmouse:
Thanks!

Issue #514, opened 2018-11-22
why DirectionsLeg is using LocalDateTime instead of ZonedDateTime?


AndreSchulzBerlin:
The time zone is lost during converstion from JSON to POJO.
For instance

> {
						"text": "06:02",
						"time_zone": "America/New_York",
						"value": 1542884540
					}

will be converted to
`LocalDateTime(chronology: IsoChronology(calendarType: iso8601, id: ISO), dayOfMonth: 22, dayOfWeek: THURSDAY, dayOfYear: 326, hour: 6, minute: 2, month: NOVEMBER, monthValue: 11, nano: 0, second: 20, year: 2018)`
when using `Gson.fromJson`

How to convert `departureTime` and `arrivalTime` to `GregorianCalendar`?

domesticmouse:
I'm happy to take a PR that keeps the timezone information.

apjanke:
Minor uh-oh: This would involve a possibly breaking public API change, since `ZonedDateTime` is not a subtype of the current `LocalDateTime`; they're alternative subtypes of `Temporal`, so it would be an incompatible change to the publicly-visible structure of the public class `DirectionsLeg`.

I think I agree with @AndreSchulzBerlin's analysis that the time zone info is being lost and this should have been a `ZonedDateTime` in the first place. I don't know what google-maps-services-java's policy/approach on potentially-breaking API changes like this is, especially since it's exposed as fields and not getter methods.

apjanke:
(Also, it probably shouldn't use `GregorianCalendar` in the API regardless; that's one of the old obsolete `java.util` time classes.)

domesticmouse:
Yes, it's a breaking change. On the up side this is a pre 1.0 library and I'd prefer the API be correct before I lock it down. 

So let's rip the band aid off now rather than later, eh? =)

apjanke:
Agreed.

apjanke:
Here's a PR: https://github.com/googlemaps/google-maps-services-java/pull/516

This affects `TransitDetails` in addition to `DirectionsLeg`.

apjanke:
PR is merged. This should be fixed now, @AndreSchulzBerlin.

PR #516, opened 2018-11-24
DirectionsLeg, TransitDetails: use ZonedDateTime instead of LocalDateTime

apjanke:
This preserves the time zone information from the API response data.

This is a breaking change to the public API! Per the discussion at https://github.com/googlemaps/google-maps-services-java/issues/514#issuecomment-441160366, this should be okay.


domesticmouse:
LGTM, thanks!

PR #513, opened 2018-11-12
Fix Google Error Prone warnings

apjanke:
This PR contains fixes for the warnings that the Google Error Prone static analyzer currently raises for the google-maps-services-java code base.

Addresses #512.

PR #517, opened 2018-11-24
Force Jacoco version 0.8.2 for OpenJDK 11 compatibility

apjanke:
The jacoco Gradle plugin seems to be defaulting to Jacoco 0.8.1, which breaks when run against OpenJDK 11. This PR adds an explicit version to force it to use 0.8.2.

JDK 11 support was added recently and picked up in Jacoco 0.8.2: https://github.com/jacoco/jacoco/issues/663

This fixes test errors as seen in https://github.com/googlemaps/google-maps-services-java/pull/513#issuecomment-439308531, which look like the following.

```
FATAL ERROR in native method: processing of -javaagent failed
Caused by: java.lang.NoSuchFieldException: $jacocoAccess
        at java.base/java.lang.Class.getField(Class.java:2000)
        at org.jacoco.agent.rt.internal_c13123e.core.runtime.ModifiedSystemClassRuntime.createFor(ModifiedSystemClassRuntime.java:138)
        ... 9 more
Process 'Gradle Test Executor 9' finished with non-zero exit value 134
org.gradle.process.internal.ExecException: Process 'Gradle Test Executor 9' finished with non-zero exit value 134
        at org.gradle.process.internal.DefaultExecHandle$ExecResultImpl.assertNormalExitValue(DefaultExecHandle.java:395)
        at org.gradle.process.internal.worker.DefaultWorkerProcess.onProcessStop(DefaultWorkerProcess.java:138)
[...]
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -88,6 +88,10 @@ javadoc {
     }
 }
 
+jacoco {
+    toolVersion = "0.8.2" // non-default version 0.8.2 required for OpenJDK 11 compatibility
+}
+
 jacocoTestReport {
     reports {
         xml.enabled = true // coveralls plugin depends on xml format report


domesticmouse:
Shiny! Thanks :-)

apjanke:
No problem!

Note for future: we can probably remove this once Gradle 5 is released and we are using it; the [default has been changed to 0.8.2 in Gradle's `master`](https://github.com/gradle/gradle/issues/6624).

domesticmouse:
Are we still barred from using gradle 5 due to the GAE libraries? I haven’t seen the warning log entries in a while, but I haven’t been looking for them either. Happy to roll up to Gradle 5rc2 to check

Issue #539, opened 2019-01-10
Add a default no-args constructor to com.google.maps.model.EncodedPolyline


clodbrasilino:
As I was using this lib for a Spring project, I faced a problem while trying to de-serialize an object from a Redis server through Spring Data Redis API. Here's the Stack Trace:

```
Failed to instantiate com.google.maps.model.EncodedPolyline using constructor NO_CONSTRUCTOR with arguments
org.springframework.data.mapping.model.MappingInstantiationException: Failed to instantiate com.google.maps.model.EncodedPolyline using constructor NO_CONSTRUCTOR with arguments 
	at org.springframework.data.convert.ReflectionEntityInstantiator.createInstance(ReflectionEntityInstantiator.java:67)
	at org.springframework.data.convert.ClassGeneratingEntityInstantiator.createInstance(ClassGeneratingEntityInstantiator.java:84)
	at org.springframework.data.redis.core.convert.MappingRedisConverter.readInternal(MappingRedisConverter.java:230)
	at org.springframework.data.redis.core.convert.MappingRedisConverter.readProperty(MappingRedisConverter.java:296)
	at org.springframework.data.redis.core.convert.MappingRedisConverter.lambda$readInternal$0(MappingRedisConverter.java:244)
	at org.springframework.data.mapping.model.BasicPersistentEntity.doWithProperties(BasicPersistentEntity.java:355)
	at org.springframework.data.redis.core.convert.MappingRedisConverter.readInternal(MappingRedisConverter.java:236)
	at org.springframework.data.redis.core.convert.MappingRedisConverter.readCollectionOrArray(MappingRedisConverter.java:791)
	at org.springframework.data.redis.core.convert.MappingRedisConverter.readProperty(MappingRedisConverter.java:283)
	at org.springframework.data.redis.core.convert.MappingRedisConverter.lambda$readInternal$0(MappingRedisConverter.java:244)
	at org.springframework.data.mapping.model.BasicPersistentEntity.doWithProperties(BasicPersistentEntity.java:355)
	at org.springframework.data.redis.core.convert.MappingRedisConverter.readInternal(MappingRedisConverter.java:236)
	at org.springframework.data.redis.core.convert.MappingRedisConverter.readCollectionOrArray(MappingRedisConverter.java:791)
	at org.springframework.data.redis.core.convert.MappingRedisConverter.readProperty(MappingRedisConverter.java:283)
[...]

PR #541, opened 2019-01-10
Add default no-arg constructor to EncodedPolyline

clodbrasilino:
As in #539


clodbrasilino:
@googlebot please check if solved

PR #543, opened 2019-01-13
Fixing Google Java Style

clodbrasilino:
Issue #539. Just fixing Google formatting.


domesticmouse:
Thanks for the fix =)

PR #565, opened 2019-05-01
Add implementation of Serializable to AutocompletePrediction

### LINKED ISSUE #563: Add implementation of Serializable into AutoCompletePrediction$MatchedSubstring class
Hello!

I am using Spring framework alongside Google Map Services and one of the features that you would like to suggest are the serialization of the class MatchedSubstring, which is located inside of the class AutoCompletePrediction. 

So far the class above mentioned does not implement Serializable, so if I try to use cache features returning data from a model that has a composition with the AutoCompletePrediction class, an error similar to the following may occur:

`Cannot serialize; nested exception is org.springframework.core.serializer.support.SerializationFailedException: Failed to serialize object using DefaultSerializer; nested exception is java.io.NotSerializableException: com.google.maps.model.AutocompletePrediction$MatchedSubstring` 

Once the class MatchedSubstring implements Serializable, it will be possible to use certain cache strategies that I am trying out. 

Thank you.

danieldsf:
This Pull Request Fixes #563.

### DIFF
--- a/src/main/java/com/google/maps/model/AutocompletePrediction.java
+++ b/src/main/java/com/google/maps/model/AutocompletePrediction.java
@@ -56,7 +56,9 @@ public class AutocompletePrediction implements Serializable {
    * Describes the location of the entered term in the prediction result text, so that the term can
    * be highlighted if desired.
    */
-  public static class MatchedSubstring {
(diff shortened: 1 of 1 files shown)


domesticmouse:
Thanks!

PR #566, opened 2019-05-01
Validate TextSearchRequest without query if type is set

### LINKED ISSUE #564: TextSearchRequest requires `query` even when it contains `type`
[Docs](https://developers.google.com/places/web-service/search#TextSearchRequests) suggest that `query` parameter becomes optional when `type` parameter is included. However this doesn't seem to work with this library. The [`validateRequest()`](https://github.com/googlemaps/google-maps-services-java/blob/master/src/main/java/com/google/maps/TextSearchRequest.java#L164) method of the `TextSearchRequest` class wants a `query` parameter in all cases that don't include a `pageToken`.

Are the docs incorrect?

danoscarmike:
Updated logic in `validateRequest` method of `TextSearchRequest` class

Fixes #564

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -10,6 +10,7 @@ Amy Boyd <amy@amyboyd.co.uk> @amyboyd
 Brantley Wells <brantleywells@gmail.com> @gitbrantley
 Brett Morgan <brettmorgan@google.com> @domesticmouse
 Chris Broadfoot <cbro@google.com> @broady
+Dan O'Meara <danom@google.com> @danoscarmike
 Dave Holmes <daveholmes@google.com> @dh--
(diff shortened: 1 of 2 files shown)


danoscarmike:
I actually started to write one but then realized there is no obvious existing test class file in which to put it. Happy to create one but I'm wondering if instead we should overload the PlacesApi.textSearchQuery method with something like:
```java
public static TextSearchRequest textSearchQuery(GeoApiContext context, PlaceType type) {
    TextSearchRequest request = new TextSearchRequest(context);
    request.type(type);
    return request;
}
```
Could easily extend existing tests then. WDYT?

domesticmouse:
I think I'll think about this after I/O =)

PR #567, opened 2019-05-02
[NOT URGENT] add new textSearchQuery (overload) method with test

danoscarmike:
This PR implements my suggestion in #566 

Adds new overload to `PlacesApi.textSearchQuery` method for cases where the user wants to pass a `type` but no `query`. This is allowable according to the [docs](https://developers.google.com/places/web-service/search#TextSearchRequests) and tests pass.


domesticmouse:
Thanks!

PR #569, opened 2019-05-05
Add Serializable definition to AutoCompletePrediction$Term class

### LINKED ISSUE #568: Add implementation of Serializable to the AutoCompletePrediction$Term class
Hello again!

Like I said in the issue #563, I am using Spring framework alongside Google Map Services and one of the features that you would like to suggest are the serialization of the class Term, which is a member of the class AutoCompletePrediction.

So far the class above mentioned does not implement Serializable, so if I try to use cache features returning data from a model that has a composition with the AutoCompletePrediction class, an error similar to the following may occur:

`Cannot serialize; nested exception is org.springframework.core.serializer.support.SerializationFailedException: Failed to serialize object using DefaultSerializer; nested exception is java.io.NotSerializableException: com.google.maps.model.AutocompletePrediction$Term`

Once the Term class implements Serializable, it will be possible to use certain cache strategies that I am trying out. This time I have tested the library within and without the change suggested and when I performed the change, the everything worked fine for me. 

Thank you.

danieldsf:
Fixes #568

### DIFF
--- a/src/main/java/com/google/maps/model/AutocompletePrediction.java
+++ b/src/main/java/com/google/maps/model/AutocompletePrediction.java
@@ -85,7 +85,9 @@ public class AutocompletePrediction implements Serializable {
    * Identifies each section of the returned description. (A section of the description is generally
    * terminated with a comma.)
    */
-  public static class Term {
+  public static class Term implements Serializable {
+
+    private static final long serialVersionUID = 1L;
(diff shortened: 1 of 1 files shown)


domesticmouse:
Thanks!

danieldsf:
You are welcome!

PR #577, opened 2019-07-03
Fix: Updated enum list of supported/unsupported PlaceType (Places API)

jmd3v:
Updated enum list of supported/unsupported PlaceType (Places API) from documentation:
https://developers.google.com/places/web-service/supported_types


jmd3v:
I signed it!

PR #578, opened 2019-07-03
Update supported/unsupported PlaceType (Places API)

jmd3v:
Removed unsupported and added some new supported PlaceType from [official docs about supported_types](https://developers.google.com/places/web-service/supported_types)
Modified one test using unsupported PlaceType.


src/main/java/com/google/maps/model/PlaceType.java
```diff
@@ -52,20 +52,15 @@
   ELECTRICIAN("electrician"),
   ELECTRONICS_STORE("electronics_store"),
   EMBASSY("embassy"),
-  ESTABLISHMENT("establishment"),
```
domesticmouse: Why are we deleting types?

jmd3v: I have eliminated those types (PlaceType) because according to the official documentation they are not valid to perform a 'nearby search' or a 'text search'.

https://developers.google.com/places/web-service/supported_types#table2

For example, 'ESTABLISHMENT' is a valid AddressType included in the PlaceDetail response but it is not supported like PlaceType in NearbySearch or TextSearch.

**Case of use:**

We first make a PlaceDetail request. The response contains one or more AddressType.
If we want to perform a 'nearby search' or 'text search' based on the types included in the PlaceDetail response. First we must transform that AddressType to a PlaceType.
But some types like ESTABLISHMENT are supported AddressType but are not PlaceType valid for Nearby / Text search.

jmd3v: Don't waste your time. Delete this pull request, because I have already make a request to service Nerby Search using ESTABLISHMENT and it works.

I do not know why the documentation says that this type is not supported in the type filter of a place search.

domesticmouse: The supported and return types are determined by systems that are beyond the control of the API team, and thus the documentation of the supported types is more indicative than definitive.

Issue #572, opened 2019-05-21
Can't retrieve status messages from Google Directions


danielolssonIT:
The DirectionResponse class doesn't have a status attribute, where errors like `MAX_ROUTE_LENGTH_EXCEEDED` can be handled. I have tried to get information via the Throwable parameter in the callback `onFailure`, but the responses there seem inconsistent. 

Is it possible to access the status message somehow?

rchopda:
static class is defined without access modifier (default) if we change the modifier to public similar to GeolocationApi Response we will be able to use google-maps-services sdk to parse the json response returned by maps api. 
Any specific reason static  api response classes are created with default access modifier

domesticmouse:
There have been several maintainers through the years, but it’s probably my fault. Please feel free to create a PR to make said class public. Thanks!

rchopda:
Hi Brett,

PR raised please validate and approve

amuramoto:
Looks like this can be closed? @domesticmouse

PR #580, opened 2019-08-01
#572: updated the access modifier for response inner classes in Api C…

rchopda:
…lasses


rchopda:
@googlebot I signed it!

PR #591, opened 2019-08-27
Making serializable PlusCode

evelzi:


evelzi:
Sure! I am working on a feature about PlaceDetails API and I have to
serialize the response of the service. When I use this API, I get an object
of type PlaceDetails that is fully serializable, except for the "plusCode"
property.  For that reason, I make the PlusCode class serializable; thus,
the response object (PlaceDetails) can be serialized without problems.

Regards!

PR #594, opened 2019-08-29
Include plus_code in PlaceDetailsRequest.FieldMask

Jasper-M:
Without it you can't request a subset of fields in a details request that includes `plus_code`.


Jasper-M:
@googlebot I signed it!

amuramoto:
Thanks for the PR!

PR #601, opened 2019-09-19
Add tourist_attraction address type

LeeU1911:
Hey,

Saw our system was failing with this new address type. So I created a PR for it. While waiting for the release, we're running off the forked version.

Thanks,
Vu


LeeU1911:
@googlebot I signed it!

PR #612, opened 2019-09-27
direct users to javadoc.io instead of github pages

### LINKED ISSUE #609: Automate javadoc publishing

jpoehnelt:
closes #609 with the simplest solution


amuramoto:
@jpoehnelt Agreed we should point to javadoc.io, but I think we should also have github.io. Since there's various docs hosted there, I think we would want users to be aware of it as a resource.

PR #625, opened 2019-10-23
Add missing Address and AddressComponent Types

adrilo:
The API started returning new types for the Address and AddressComponents. These missing types generate warnings.


adrilo:
@googlebot I signed it!

PR #627, opened 2019-10-23
Add missing Address and AddressComponent Types

adrilo:
The API started returning new types for the Address and AddressComponents. These missing types generate warnings.


jpoehnelt:
FYI your CLA is for a different email, feel free to sign and comment again.

Regarding the pr, thanks for catching these types. I still need to dig in a bit deeper but there seems to be some confusion between place types, address component, and gson types going on. I'm going to attempt to find the source of truth here and update to match.

PR #629, opened 2019-10-31
remove AddressType and update PlaceType and AddressComponentType

jpoehnelt:
This fixes the misuse of AddressType where PlaceType should be used.


Abosnd02:
35177729+Abosnd02@users.noreply.github.com

Issue #635, opened 2019-12-12
Google Maps fails to use TLS 1.2 in Java 8 environment


bmccauley:
I am running Google Maps 0.10.1 under Java 1.8.0_221. When I try to run the given sample code:

```
GeoApiContext context = new GeoApiContext.Builder()
    .apiKey("API_KEY_HERE")
    .build();
 GeocodingResult[] results =  GeocodingApi.geocode(context,
    "1600 Amphitheatre Parkway Mountain View, CA 94043").await();
Gson gson = new GsonBuilder().setPrettyPrinting().create();
System.out.println(gson.toJson(results[0].addressComponents));
```

I get an exception:

```
java.net.UnknownServiceException: Unable to find acceptable protocols. isFallback=false, modes=[ConnectionSpec(cipherSuites=[TLS_AES_128_GCM_SHA256, TLS_AES_256_GCM_SHA384, TLS_CHACHA20_POLY1305_SHA256, TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256, TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256, TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384, TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384, TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256, TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256, TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA, TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA, TLS_RSA_WITH_AES_128_GCM_SHA256, TLS_RSA_WITH_AES_256_GCM_SHA384, TLS_RSA_WITH_AES_128_CBC_SHA, TLS_RSA_WITH_AES_256_CBC_SHA, SSL_RSA_WITH_3DES_EDE_CBC_SHA], tlsVersions=[TLS_1_3, TLS_1_2], supportsTlsExtensions=true), ConnectionSpec()], supported protocols=[TLSv1]
```

Google Maps believes that TLSv1 is the only supported protocol. However, if I test this by creating an SSLSocket:

```
SSLSocketFactory socketFactory = (SSLSocketFactory) SSLSocketFactory.getDefault();
SSLSocket sslSocket = (SSLSocket) socketFactory.createSocket("www.verisign.com", 443);
[...]

PR #636, opened 2019-12-13
fix: upgrade okhttp to 4.2.2

jpoehnelt:
#635 might be fixed by this. need to confirm


giakhangdouble:
Người đi tìm công  nghệ
8p6xwa

PR #638, opened 2019-12-20
style: Update pull request template.

arriolac:
Add a checklist when users make a PR.


jpoehnelt:
Actually, can you leave the label in.

arriolac:
> Actually, can you leave the label in.

Yep, let me add back (same comment across other PRs).

PR #639, opened 2019-12-29
Fix `MaxRouteLengthExceededException` documentation

Stephan202:
The original documentation appears to have been copied from `MaxElementsExceededException`.


amuramoto:
@Stephan202 Thanks for another PR!

PR #642, opened 2020-01-02
documentation problem

ironblaster:
in the readme file, the call of the shutdown () method is not specified at the end of a request.

this is a problem because after n requests java creates multiple threads not closed
and in the case of tomcat the instance goes into error

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,8 +193,6 @@ life of your application. This will enable proper QPS enforcement across all of 
 At the end of the execution, call the `shutdown()` method of `GeoApiContext`,
 otherwise the thread will remain instantiated in memory.
 
-
-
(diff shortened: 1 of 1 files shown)


ironblaster:
@googlebot I signed it!

README.md
```diff
@@ -139,7 +139,7 @@ You can find the latest version at the top of this README or by [searching
 Maven Central](https://search.maven.org/#search%7Cga%7C1%7Ca%3A%22google-maps-services%22) or [Gradle, Please](http://gradleplease.appspot.com/#google-maps-services).
 
 ## Developer Documentation
-
+
```
jpoehnelt:
```suggestion

```

ironblaster:
thanks, do I have to do more to confirm?

PR #644, opened 2020-01-06
feat: add support for experience id

jpoehnelt:
- [x] add tests
- [x] add setter on client since experience id will change often


arriolac:
Will reopen so I can request a review from you instead.

PR #378, opened 2017-11-13
AddressType: add beauty care types

apjanke:
Fixes #279 (from comment https://github.com/googlemaps/google-maps-services-java/issues/279#issuecomment-339551783).

I can reproduce that comment's warnings locally, and this patch fixes them.

PR #646, opened 2020-01-09
Added primary school as a poi AddressType to silence SafeEnumAdapter …

jdcohen220:
This is similar to  #377 and #378. Added the primary school element to AddressType enums


jdcohen220:
@googlebot I signed it!

PR #650, opened 2020-01-17
fixed: Added path exception when there is no center or no zoom

danieljesus:
We saw in the documentation (https://developers.google.com/maps/documentation/maps-static/dev-guide?hl=es#ImplicitPositioning ) these lines:

"Normally, you need to specify center and zoom URL parameters to define the location and zoom level of your generated map. However, if you supply markers, path, or visible parameters, you can instead let the Maps Static API determine the correct center and zoom level implicitly, based on evaluation of the position of these elements."


amuramoto:
@danieljesus  Thanks for the PR!

PR #654, opened 2020-01-22
Typo in FieldMask fixed

cezarykluczynski:
This is a fix for field `GEOMETRY_LOCATION_LNG` from `FieldMask` enum. Google Place API was unable to recognize `lon`. After changing to `lng` it works fine.


cezarykluczynski:
@googlebot I signed it!

PR #656, opened 2020-01-27
build(Gradle): upgrade gradle wrapper to 6.1.1

arriolac:
- Upgrading Gradle wrapper to 6.1.1
- Adding `withJavadocJar` & `withJavadocJar` to ensure that javadoc + sources .jar are published when running `./gradlew publish`

I verified that this works when manually running `./gradlew publish`. Still need to verify that this works with the deploy script that travis executes (we can verify this on the next release)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -86,6 +86,11 @@ tasks.withType(Test) {
     systemProperty 'client.secret', System.getProperty('client.secret')
 }
 
+java {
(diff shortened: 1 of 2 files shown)


arriolac:
Reran the Travis build a few times and now it passes. Haven't dug into why it's so finicky but failures seems to be unrelated to this change.

Issue #663, opened 2020-03-03
Add monitoring metrics for latency and errors


yangd102:
**Is your feature request related to a problem? Please describe.**
It would be good to have monitoring metrics that would allow us to see the latency and error rates of API methods.

 **Describe the solution you'd like**
Create hooks into the SDK to allow somebody to observe latency and error rates. By default make the hooks do nothing. Provide an OpenCensus plugin that would export the latency and error rates - eg to Datadog or Cloud Stackdriver etc.

PR #665, opened 2020-03-04
feat: Add OpenCensus implementation for performance logging

yangd102:
This is an OpenCensus implementation for RequestMetrics added in #664.
By default it OpenCensus will not log to any target destination.

This is providing a useful implementation for #663 🦕

### DIFF
--- a/src/main/java/com/google/maps/metrics/OpenCensusRequestMetrics.java
+++ b/src/main/java/com/google/maps/metrics/OpenCensusRequestMetrics.java
@@ -1,0 +1,75 @@
+package com.google.maps.metrics;
+
+import io.opencensus.stats.StatsRecorder;
+import io.opencensus.tags.TagContext;
+import io.opencensus.tags.TagValue;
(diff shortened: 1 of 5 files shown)


yangd102:
I will write an additional test for the census part, however the actual code will remain unchanged.

src/main/java/com/google/maps/metrics/OpenCensusRequestMetrics.java
```diff
@@ -0,0 +1,75 @@
...
+  private String exceptionName(Exception exception) {
+    if (exception == null) {
+      return "";
+    } else {
+      return exception.getClass().getName();
+    }
+  }
+
+  private long milliTime() {
+    return System.nanoTime() / 1000000l;
```
arriolac:
```suggestion
    return System.currentTimeMillis();
```

yangd102:
PTAL, I added a test for OpenCensus and applied suggestion.

arriolac:
LGTM

PR #676, opened 2020-04-22
feat: Add business_status

arriolac:
Adding business_status as a field to the place details and place search responses. Also added a new field mask for the place details request.


protonforest:
Hi. Currently waiting on this to be merged in :)

PR #677, opened 2020-04-22
build: Adding build, release, and publish workflows.

arriolac:
Adding GitHub workflows.


.github/workflows/build.yml
```diff
@@ -0,0 +1,40 @@
...
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    - name: Run tests
+      run: |
+        echo "Running unit tests"
+        ./gradlew check test jacocoTestReport --stacktrace
+
+        echo "Sending test results to CodeCov"
+        bash <(curl -s https://codecov.io/bash)
```
jpoehnelt, line 40: maybe put these in separate steps?

PR #692, opened 2020-07-20
refactor: Changed exception test from try/catch block to ExpectedException rule

eas5:
**Problem:**
Explicitly catching or throwing exceptions in test methods may result in flaky behavior if an unexpected exception occurs.

**Solution:**
JUnit4.x provides the ExpectedException rule to test for exceptions and its messages in a straightforward way.

**Result:**
_Before:_
    try {
      builder.build().get(new ApiConfig("/"), GeocodingApi.Response.class, "k", "v").await();
    } catch (IOException ioe) {
      // Ensure the message matches the status line in the mock responses.
      assertEquals("Server Error: 500 Internal server error", ioe.getMessage());
      return;
    }
    fail("Internal server error was expected but not observed.");

_After:_
    // Configure rule to the exception that must be thrown and ensure 
    // the message matches the status line in the mock responses 
    thrown.expect(IOException.class);
    thrown.expectMessage("Server Error: 500 Internal server error");
    builder.build().get(new ApiConfig("/"), GeocodingApi.Response.class, "k", "v").await();

### DIFF
--- a/src/test/java/com/google/maps/GeoApiContextTest.java
+++ b/src/test/java/com/google/maps/GeoApiContextTest.java
@@ -42,14 +42,19 @@ import okhttp3.mockwebserver.MockWebServer;
 import okhttp3.mockwebserver.RecordedRequest;
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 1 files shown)


eas5:
@googlebot I signed it!

PR #710, opened 2021-01-18
chore(deps): Upgrade Gson 2.8.5 -> 2.8.6

hisener:
See https://github.com/google/gson/compare/gson-parent-2.8.5...gson-parent-2.8.6

---

This causes [dependency convergence](https://maven.apache.org/enforcer/enforcer-rules/dependencyConvergence.html) in downstream code since 2.8.5 is a very old version.


hisener:
@googlebot I signed it!

PR #722, opened 2021-02-16
fix: allow using existing session token(String) in place autocomplete

### LINKED ISSUE #689: Allow using existing session token(String) in place autocomplete
In our case the session token is not generated on the service that does the Google Places API call. Instead an existing token is passed from the client to facilitate correct billing. The token is not necessarily a UUID and there is no reason why only UUID should be allowed as implemented here #478. The API accepts any string as a token.

My request is to add another constructor to `SessionToken`:
```java
public SessionToken(String token) {
  this.token = token;
}
```

SethMyers:
Fixes #689

### DIFF
--- a/src/main/java/com/google/maps/PlaceAutocompleteRequest.java
+++ b/src/main/java/com/google/maps/PlaceAutocompleteRequest.java
@@ -68,6 +68,15 @@ public class PlaceAutocompleteRequest
     }
 
     /**
+     * Construct a session that is a continuation of a previous session.
+     *
+     * @param token The unique token to be used as the identifier for this session.
+     */
+    public SessionToken(String token) {
+      this.uuid = UUID.nameUUIDFromBytes(token.getBytes());
+    }
+
+    /**
      * Retrieve the universally unique identifier for this session. This enables you to recreate the
      * session token in a later context.
      *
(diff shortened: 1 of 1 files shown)


SethMyers:
@googlebot I signed it!

src/main/java/com/google/maps/PlaceAutocompleteRequest.java
```diff
@@ -67,6 +67,15 @@ public SessionToken(UUID uuid) {
       this.uuid = uuid;
     }
 
+    /**
+     * Construct a session that is a continuation of a previous session.
+     *
+     * @param token The unique token to be used as the identifier for this session.
```
arriolac, line 73: I would expect `token` here to be a string representation of a UUID after reading this comment, but that's not the case given the implementation. While this provides convenience, I think this is potentially error-prone. I suggest either clarifying the comment, or keeping the existing API of constructing a `SessionToken` from a `UUID`. The latter seems like a clearer API to me, but if a convenience initializer is useful for others, I'm happy to merge in.

arriolac (changes requested):

src/main/java/com/google/maps/PlaceAutocompleteRequest.java
SethMyers: would
`@param token The character string to be used as the unique identifier for this session.`
be clear enough?

vojtapol: > keeping the existing API
The existing API does not work well if the session identifier is not a UUID.

arriolac (approved):

PR #723, opened 2021-02-17
Feat: Add LONG_DISTANCE_TRAIN to VehicleType and fallback to OTHER if unknown value

### LINKED ISSUE #718: Directions: Missing VehicleType LONG_DISTANCE_TRAIN
The vehicle type "LONG_DISTANCE_TRAIN" (as described [here](https://developers.google.com/maps/documentation/directions/overview#VehicleType)) is missing from the [VehicleType](https://github.com/googlemaps/google-maps-services-java/blob/master/src/main/java/com/google/maps/model/VehicleType.java) enum.

This will result in the `type` attribute beeing `null` inside the [Vehicle](https://github.com/googlemaps/google-maps-services-java/blob/master/src/main/java/com/google/maps/model/Vehicle.java).

I would further suggest to add a `SafeEnumAdapter` for the `VehicleType` in the [OkHttpPendingResult](https://github.com/googlemaps/google-maps-services-java/blob/a78722087bff37d36b403b853ce28517c32037ac/src/main/java/com/google/maps/internal/OkHttpPendingResult.java#L266) and the [GaePendingResult](https://github.com/googlemaps/google-maps-services-java/blob/a78722087bff37d36b403b853ce28517c32037ac/src/main/java/com/google/maps/internal/GaePendingResult.java#L188).

LumnitzF:
Fixes #718 🦕

### DIFF
--- a/src/main/java/com/google/maps/internal/GaePendingResult.java
+++ b/src/main/java/com/google/maps/internal/GaePendingResult.java
@@ -41,6 +41,7 @@ import com.google.maps.model.OpeningHours.Period.OpenClose.DayOfWeek;
 import com.google.maps.model.PlaceDetails.Review.AspectRating.RatingType;
 import com.google.maps.model.PriceLevel;
 import com.google.maps.model.TravelMode;
+import com.google.maps.model.VehicleType;
 import java.io.IOException;
 import java.nio.charset.Charset;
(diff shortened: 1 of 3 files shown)


arriolac (approved):
This is great! Thanks for the PR.

